PR #10, opened 2016-11-30
Fix diff for multi option custom field with different order of option…

cdworak:
…s to return true

### DIFF
--- a/customfield.go
+++ b/customfield.go
@@ -106,7 +106,7 @@ func (m MultiOption) CustomFieldTag() string {
 }
 
 func (m MultiOption) Equal(c CustomFieldValueComparable) bool {
-	n := *c.(*MultiOption)
+	n := c.(MultiOption)
 	if len(m) != len(n) {
 		return false
 	}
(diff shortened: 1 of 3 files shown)


customfield.go
```diff
@@ -101,6 +105,26 @@ func (m MultiOption) CustomFieldTag() string {
 	return CUSTOMFIELDTYPE_MULTIOPTION
 }
 
+func (m MultiOption) Equal(c CustomFieldValueComparable) bool {
+	n := c.(MultiOption)
+	if len(m) != len(n) {
+		return false
+	} else {
```
mhupman: I'd pull this out of the else, its obvious that its next.

customfield.go
```diff
@@ -101,6 +105,26 @@ func (m MultiOption) CustomFieldTag() string {
...
+		for i := 0; i < len(m); i++ {
+			found := false
+			for j := 0; j < len(n); j++ {
+				if m[i] == n[j] {
+					found = true
+				}
+			}
+			if found != true {
```
mhupman: `if !found`?

mhupman:
@benmcginnis Can you peep this fella too?

PR #13, opened 2017-01-19
Allow Labels to not maintain Order

benmcginnis:


location_diff.go
```diff
@@ -51,7 +51,17 @@ func (y Location) Diff(b *Location) (d *Location, diff bool) {
 			continue
 		}
 
-		if !reflect.DeepEqual(valA.Interface(), valB.Interface()) {
+		aI, bI := valA.Interface(), valB.Interface()
+
+		unorderedA, aOk := aI.(*UnorderedStrings)
```
mhupman: Shouldn't this just check if the object satisfies an Equal()/Compare() interface?  That way we don't have to add an exception for every type....

mhupman:
Can we figure out how to share the compare logic between the multioption equals and this one?  Seems sad they are basically dupes of each other.

location.go
```diff
@@ -467,3 +467,30 @@ type HolidayHours struct {
 	Date  string `json:"date"`
 	Hours string `json:"hours"`
 }
+
+// UnorderedStrings masks []string properties for which Order doesn't matter, such as LabelIds
+type UnorderedStrings []string
+
+// Equal implements compares UnorderedStrings
```
mhupman: grammar

location_diff.go
```diff
@@ -51,7 +55,17 @@ func (y Location) Diff(b *Location) (d *Location, diff bool) {
 			continue
 		}
 
-		if !reflect.DeepEqual(valA.Interface(), valB.Interface()) {
+		aI, bI := valA.Interface(), valB.Interface()
+
+		comparableA, aOk := aI.(Comparable)
+		comparableB, bOk := bI.(Comparable)
+
+		if aOk && bOk {
```
mhupman: Can we just inline these checks?  The number of local scoped variables with A and B is confusing

benmcginnis: I don't think the magical inlining works with two type checks:
```
comparableA, aOk, comparableB, bOk := aI.(Comparable), bI.(Comparable)
```
gives "assignment count mismatch"

and this
```
if comparableA, aOk:= aI.(Comparable); comparableB, bOk := bI.(Comparable); aOk && bOk {
```
gives `syntax error missing { after if clause`

PR #16, opened 2017-02-07
Update SpecialOffer and Lists to use new fields

benmcginnis:


location.go
```diff
@@ -41,8 +44,8 @@ type Location struct {
 	MobilePhone            *string                `json:"mobilePhone,omitempty"`
 	TollFreePhone          *string                `json:"tollFreePhone,omitempty"`
 	TtyPhone               *string                `json:"ttyPhone,omitempty"`
-	SpecialOffer           *string                `json:"specialOffer,omitempty"`
-	SpecialOfferUrl        *string                `json:"specialOfferUrl,omitempty"`
+	SpecialOffer           *string                `json:"featuredMessage,omitempty"`
```
mhupman: Should we update the names of the types?

benmcginnis: Sure.  My only thought was that client code wouldn't have to update just because the json key changed.

mhupman: Yeah I think your motivation is correct, but nobody is using it yet :) Lets keep it clean.

PR #18, opened 2017-03-10
V2

cdworak:


error.go
```diff
@@ -62,3 +62,18 @@ func (e Errors) Warnings() []Error {
 	}
 	return warnings
 }
+
+func IsNotFoundError(err error) bool {
+	if e, ok := err.(Errors); ok {
+		for _, innerError := range e {
+			if IsNotFoundError(innerError) {
```
mhupman: Didn't we find this needed to be a pointer?

error_test.go
```diff
@@ -0,0 +1,56 @@
+package yext
+
+import "testing"
+
+func TestIsNotFoundError(t *testing.T) {
```
mhupman: yay!

location_diff.go
```diff
@@ -22,7 +26,8 @@ import "reflect"
 //   // isDiff -> false
 //   // delta -> nil
 func (y Location) Diff(b *Location) (d *Location, diff bool) {
-	// TODO(bjm): remove this once the ETL is upgraded to use hydration
+	// TODO (bjm):
+	// remove this once the ETL is upgraded to use hydration
```
mhupman: why the newlines?

user_diff_test.go
```diff
@@ -3,7 +3,7 @@ package yext_test
 import (
 	"testing"
 
-	"github.com/yext/yext-go"
+	yext "gopkg.in/yext/yext-go.v2"
```
mhupman: Shouldn't need to import at all since we're in the same package...

error.go
cdworak: having it be a pointer was what was causing it to both skip the  if (Errors), else if (Error) when we passed the single innerError. When I removed the pointer the tests all pass. Adding the pointer in fails the single Error tests

mhupman: Oh, ok.  Was confused.

user_diff_test.go
cdworak: it's in package yext_test...i'm not sure why considering all the other tests (ex: customfield_test) are in package yext. i'll move to package yext unless there's any reason not to

PR #22, opened 2017-06-23
added page token param J=PC-18626

dnl314:


client.go
```diff
@@ -308,6 +314,9 @@ func addListOptions(requrl string, opts *ListOptions) (string, error) {
 	if opts.Offset != 0 {
 		q.Add("offset", strconv.Itoa(opts.Offset))
 	}
+	if opts.PageToken != "" {
```
mhupman: I think we should only allow PageToken _or_ Offset, and probably prefer PageToken

location_service.go
```diff
@@ -38,6 +39,7 @@ func (l *LocationService) ListAll() ([]*Location, error) {
 	var lg listRetriever = func(opts *ListOptions) (int, int, error) {
 		llo.ListOptions = *opts
 		llr, _, err := l.List(llo)
+		llo.ListOptions.PageToken = llr.NextPageToken
```
mhupman: I think its odd that we mix a direct set of PageToken here, but the return values in listRetriever for Count support.  This also means that other "list" endpoints would need to be updated manually instead of this being automatic.  I think we should either generify or come up with a "page token" listRetriever variant.

client.go
```diff
@@ -270,17 +272,21 @@ func listHelper(lr listRetriever, opts *ListOptions) error {
...
+			if reportedtotal <= opts.Offset+opts.Limit {
+				break
+			}
+			opts.Offset += opts.Limit
 		}
-
-		opts.Offset += opts.Limit
 	}
 
 	// Safety check
```
mhupman: If we're in PageToken mode, then we don't/shouldn't need the count validation.

client.go
```diff
@@ -291,6 +290,31 @@ func listHelper(lr listRetriever, opts *ListOptions) error {
...
+	for {
+		nextpagetoken, err := lr(opts)
+		if err != nil {
+			return err
+		}
+
+		if nextpagetoken != "" {
+			opts.PageToken = nextpagetoken
+			hasPrevPageToken = true
+		} else if hasPrevPageToken {
```
mhupman: I'm not sure I understand the boolean... can we just break in a `else` (where nextpagetoken == "")?

client.go
```diff
@@ -305,7 +329,9 @@ func addListOptions(requrl string, opts *ListOptions) (string, error) {
 	if opts.Limit != 0 {
 		q.Add("limit", strconv.Itoa(opts.Limit))
 	}
-	if opts.Offset != 0 {
+	if opts.PageToken != "" {
```
mhupman: nice - like either or behavior!

location_service_test.go
```diff
@@ -120,90 +117,67 @@ func TestListAll(t *testing.T) {
...
+		client.LocationService.ListBySearchIds([]string{"1234"})
+		if reqs < len(test.expectedTokenRequests) {
+			t.Errorf("Too few requests sent to location list - got %d, expected %d", reqs, len(test.expectedTokenRequests))
 		}
 
 		teardown()
 	}
 }
 
-func TestListMismatchCount(t *testing.T) {
```
mhupman: This test should not go away... you might have to change it to use something other than LocationService, though

mhupman: (might need to move it to another service_test file)

location_service_test.go
```diff
@@ -85,33 +83,32 @@ func makeLocs(n int) []*Location {
 	return locs
 }
 
-func TestListAll(t *testing.T) {
```
mhupman: same comments as the mismatch test... we want this somewhere, probably need to move it though

PR #23, opened 2017-06-29
J=PC-18560 Update searchid query param

xiexieeric:


mhupman:
Builds failing!  After fixing, don't need to attribute the JIRA ticket here, and make sure to squash pre-merge.

PR #25, opened 2017-07-25
Add GoogleAttributes to Location

cdworak:


location.go
```diff
@@ -631,3 +635,73 @@ func (a *UnorderedStrings) Equal(b Comparable) bool {
...
+	}
+
+	var (
+		u = GoogleAttribute(*a)
+		s = GoogleAttribute(*b.(*GoogleAttribute))
+	)
+	if *u.Id != *s.Id {
+		return false
+	}
+	for i := range *u.OptionIds {
```
benmcginnis: shouldn't you check the lengths and return false if they're not the same?

location.go
```diff
@@ -563,6 +560,13 @@ func (y Location) GetAdmittingHospitals() (v []string) {
 	return v
 }
 
+func (y Location) GetGoogleAttributes() []*GoogleAttribute {
```
benmcginnis: should this return the type or do we want it to return the raw array?

location.go
```diff
@@ -631,3 +635,73 @@ func (a *UnorderedStrings) Equal(b Comparable) bool {
...
+
+	var (
+		u = GoogleAttribute(*a)
+		s = GoogleAttribute(*b.(*GoogleAttribute))
+	)
+	if *u.Id != *s.Id {
+		return false
+	}
+	for i := range *u.OptionIds {
+		if (*u.OptionIds)[i] != (*s.OptionIds)[i] {
```
benmcginnis: also, does order matter? I think this is making it so that order would matter

location_diff_test.go
```diff
@@ -373,6 +373,104 @@ func TestBoolDiffs(t *testing.T) {
...
+			&GoogleAttribute{Id: String("has_catering"), OptionIds: Strings([]string{"false"})},
+		},
+		newValue: &GoogleAttributes{
+			&GoogleAttribute{Id: String("has_catering"), OptionIds: Strings([]string{"false"})},
+			&GoogleAttribute{Id: String("has_delivery"), OptionIds: Strings([]string{"true"})},
+		},
+		isDiff: true,
+		expectedFieldValue: GoogleAttributes{
+			&GoogleAttribute{Id: String("has_catering"), OptionIds: Strings([]string{"false"})},
+			&GoogleAttribute{Id: String("has_delivery"), OptionIds: Strings([]string{"true"})},
```
benmcginnis: can we get a larger list of google attributes to test?

benmcginnis: like maybe ones that aren't just true/false?

PR #28, opened 2017-08-18
Add comparison for LocationList type

benmcginnis:


mhupman:
Well, actually, would love a unit test, but ok :(

PR #36, opened 2018-01-11
Added some review fields and fixed a few location ones

tbirdsall:


review.go
```diff
@@ -128,14 +143,14 @@ func (y Comment) GetId() int {
 }
 
 func (y Comment) GetParentId() int {
-	if y.Id != nil {
+	if y.ParentId != nil {
```
bhainesva: ☠️

PR #37, opened 2018-01-17
Nil slice failing test

bhainesva:


bhainesva:
Not sure what's up with Travis, I don't think it's my tests.

mhupman:
@bhainesva Seems like an import of yext-go inside yext-go?

```
import cycle not allowed in test
package github.com/yext/yext-go (test)
	imports github.com/yext/yext-go
FAIL	github.com/yext/yext-go [setup failed]
```

location_diff.go
```diff
@@ -101,11 +101,15 @@ func (y Location) Diff(b *Location) (d *Location, diff bool) {
 	return d, diff
 }
 
+// isNil determines if an interface's underlying data is nil
+func isNil(a interface{}) bool {
+	defer func() { recover() }()
```
mhupman: yikes!  what is this for?

location_diff.go
```diff
@@ -47,7 +47,7 @@ func (y Location) Diff(b *Location) (d *Location, diff bool) {
 			continue
 		}
 
-		if valB.IsNil() || nameA == "Id" {
+		if valB.IsNil() || isNil(getUnderlyingValue(valB)) || nameA == "Id" {
```
mhupman: Interesting... I feel like we should be able to take advantage of IsNil() for the nil check.  Maybe we need to two helpers:

`unwrapPointer(reflect.Value) reflect.Value` that loops until Kind() is not Ptr or Interface
and
`getUnderlyingValue(reflect.Value) interface{}` that calls unwrapPointer() and then returns the .Interface()

Thoughts?

location_diff.go
bhainesva: If ValueOf(a) doesn't have specific kinds then IsNil will panic. I guess I could just add a check for all of the valid ones and return false if it's not valid. 

https://stackoverflow.com/questions/13476349/check-for-nil-and-nil-interface-in-go

location_diff.go
bhainesva: I tried doing something like but was having trouble getting it to work. I'll keep looking at it.

bhainesva: @mhupman I removed the recover but I can't figure out how to avoid some kind of Kind switch like this. `.IsNil()` isn't generally safe to call on any Value. It works for `valB.IsNil()` because all of the fields on a Location are pointers but if we had a pure `string` it would panic. Once we start unwrapping pointers to check the real values we can't use `.IsNil()` anymore.

mhupman: I can't think of a better way to do it.  lgtm

PR #38, opened 2018-01-17
Added analytics service J=PC-23360

tbirdsall:


analytics_data.go
```diff
@@ -0,0 +1,412 @@
+package yext
+
+type AnalyticsData struct {
+	ProfileViews                          *int     `json:"Profile Views"`
```
mhupman: do we need the JSON tags?  They seem funky

analytics_service.go
```diff
@@ -0,0 +1,69 @@
...
+	FoursquareCheckinGender    *string   `json:"foursquareCheckinGender"`
+	FoursquareCheckinTimeOfDay *string   `json:"foursquareCheckinTimeOfDay"`
+	InstagramContentType       *string   `json:"instagramContentType"`
+	Age                        *[]string `json:"age"`
+	Gender                     *string   `json:"gender"`
+	FacebookImpressionType     *[]string `json:"facebookImpressionType"`
+	FacebookStoryType          *[]string `json:"facebookStoryType"`
+}
+
+type AnalyticsReportBody struct {
```
mhupman: AnalyticsReportRequest?

analytics_service.go
```diff
@@ -0,0 +1,69 @@
...
+	Dimensions []string          `json:"dimensions"`
+	Filters    *AnalyticsFilters `json:"filters"`
+}
+
+type AnalyticsReportResponse struct {
+	Data []*AnalyticsData `json:"data"`
+	Id   int              `json:"id"`
+}
+
+func (l *AnalyticsService) Create(jsonData *AnalyticsReportBody) (*AnalyticsReportResponse, *Response, error) {
```
mhupman: prefer `a`, `arr`, `req` as argument name.  its not actually JSON.

analytics_data.go
tbirdsall: Hmm, not quite sure - I was not able to get the api call to work without it. What do you mean by "funky"?

mhupman: They have spaces in them

tbirdsall: Yep! That's purposeful, that's how they are in the response (verified in Postman and dev logs). Interestingly, all the report data based on metrics are uppercase with spaces in between, but the ones based on dimensions are lowercase and camelcased

mhupman: Interesting, cool!

PR #40, opened 2018-01-22
location_diff: if custom field missing, don’t assume delta

cdworak:
Not sure how to test yet, but for visibility I think this is the last of the duplicate updates.
As we already new, if the custom field is not filled out it is missing from the API response. the diff assumes that if the custom field is not there that it is a delta...but if the value is the nil value then there should be no delta.

You can see this in the pf changs database when we've been sending "" daily for single line text fields (p.f. chang's is ignore history so it compares against the API).

### DIFF
--- a/location.go
+++ b/location.go
@@ -159,10 +159,6 @@ type Location struct {
 	*/
 }
 
-func (y Location) GetNilIsEmpty() bool {
-	return y.nilIsEmpty
-}
-
 func (y Location) GetId() string {
 	if y.Id != nil {
 		return *y.Id
--- a/location.go
+++ b/location.go
@@ -159,6 +159,10 @@ type Location struct {
 	*/
 }
 
+func (y Location) GetNilIsEmpty() bool {
+	return y.nilIsEmpty
+}
+
 func (y Location) GetId() string {
 	if y.Id != nil {
 		return *y.Id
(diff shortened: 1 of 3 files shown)


location.go
```diff
@@ -159,6 +159,10 @@ type Location struct {
 	*/
 }
 
+func (y Location) GetNilIsEmpty() bool {
```
mhupman: why do we need/want a public accessor?

location_diff.go
```diff
@@ -88,7 +88,7 @@ func (y Location) Diff(b *Location) (d *Location, diff bool) {
 							diff = true
 							d.CustomFields[field] = value
 						}
-					} else {
+					} else if !isZeroValue(reflect.ValueOf(value), b.GetNilIsEmpty()) {
```
mhupman: Hmm.  @benmcginnis does this mess w/ your find re: photos?

mhupman (changes requested):
Gotta have unit tests with this

location_diff_test.go
```diff
@@ -962,6 +962,105 @@ func TestCustomFieldsDiff(t *testing.T) {
...
+	},
+}
+
+func (t customFieldsNilIsEmptyTest) formatErrorBase(index int) string {
+	return fmt.Sprintf("Failure with example %v:\n\tbase: '%v'\n\tnew: '%v'", index, t.baseValue, t.newValue)
+}
+
+func TestCustomFieldsNilIsEmptyDiff(t *testing.T) {
+	a, b := *new(Location), new(Location)
+	for i, data := range customFieldsNilIsEmptyTests {
```
mhupman: `test` instead of `data`?

mhupman (approved):
nice!  love the tests :)

PR #42, opened 2018-01-24
allow for typed nils for struct type custom fields

benmcginnis:


mhupman (approved):
could you add some simple tests for getUnderlyingValue since that is the real thing being tested here?  you can leave the "higher" level test you put in place, but I'd also like to ensure that:

`reflect.DeepEqual(getUnderlyingValue((*yext.Photo)(nil)), (*yext.Photo)(nil))`

if possible

mhupman (approved):
awesome!

PR #43, opened 2018-01-25
J=PC-24642 updated mustcachecustomfields and added a test for it

byronharvey24:


customfield_service.go
```diff
@@ -330,10 +330,11 @@ func (s *CustomFieldService) CacheCustomFields() error {
 	return nil
 }
 
-func (s *CustomFieldService) MustCacheCustomFields() {
+func (s *CustomFieldService) MustCacheCustomFields() []*CustomField {
```
mhupman: Can CacheCustomFields() be updated to return the slice of custom fields too?  Then this function can just return that directly.

mhupman (approved):
one tweak, but lgtm.  nice unit test!

customfield_service.go
```diff
@@ -320,20 +320,22 @@ func (s *CustomFieldService) List(opts *ListOptions) (*CustomFieldResponse, *Res
 	return v, r, nil
 }
 
-func (s *CustomFieldService) CacheCustomFields() error {
+func (s *CustomFieldService) CacheCustomFields() (error, []*CustomField) {
```
mhupman: `error` is always the last return value in Go

mhupman (approved):
one more thing, but lgtm!

PR #44, opened 2018-01-30
location: add constants for hours

cdworak:


mhupman (approved):
nice good stuff!  Can you update the HoursIsEquivalent() thing to use the closed constant (or did I not use this string?)

PR #46, opened 2018-03-01
Hours Helper

cdworak:
Let me know if this was along the lines of what you were envisioning! Wasn't in the sprint but built it along with the taco bell export.


benmcginnis (approved):
awesome! lgtm

customfield_service.go
```diff
@@ -616,6 +616,8 @@ func (c *CustomFieldManager) GetStringSlice(name string, loc *Location) ([]strin
 		return []string(*fv.(*TextList)), nil
 	case *LocationList:
 		return []string(*fv.(*LocationList)), nil
+	case MultiOption:
```
mhupman, line 619: Why is LocationList a pointer but this isn't?

hours.go
```diff
@@ -0,0 +1,194 @@
...
+const (
+	HoursClosedAllWeek = "1:closed,2:closed,3:closed,4:closed,5:closed,6:closed,7:closed"
+	HoursOpen24Hours   = "00:00:00:00"
+	HoursClosed        = "closed"
+)
+
+type Weekday int
+
+const (
+	Sunday Weekday = iota + 1
```
mhupman: Could you add a note or link to documentation here around _why_ we follow this indexing pattern?

hours.go
```diff
@@ -0,0 +1,194 @@
...
+		return h.Thursday
+	case Friday:
+		return h.Friday
+	case Saturday:
+		return h.Saturday
+	}
+	return ""
+}
+
+func (h *HoursHelper) Serialize() string {
```
mhupman: exactly the function name I was thinking!

mhupman (approved):
* are we supporting split hours?  do we need to?
* Are the tests comprehensive enough given the current behavior of the API?  I just can't recall.  Do we always get all day indices?

customfield_service.go
cdworak: lol beats me. the multi option has always not been a pointer but i can add a case to check for it just in case though if you try to case to a *yext.MultiOption after doing a custom field cache you'll get an error

cdworak: Sorry, if you look above the lines included in the patch you'll see the cases for non pointer (for LocationList, for example)

hours.go
cdworak: Yes! I read item descriptions sometimes :)

hours.go
```diff
@@ -89,7 +101,34 @@ func (h *HoursHelper) SetHours(weekday Weekday, hours string) {
...
+	case Saturday:
+		h.Saturday = append(h.Saturday, hours)
+	}
+}
+
+func (h *HoursHelper) SetClosed(weekday Weekday) {
+	h.SetHours(weekday, []string{HoursClosed})
+}
+
+func (h *HoursHelper) SetUnspecified(weekday Weekday) {
```
mhupman, line 127: oh nice

mhupman (approved):
lgtm

PR #51, opened 2018-04-16
Add case to handle case when custom field (single option) is blank

cdworak:


customfield_service.go
```diff
@@ -580,8 +580,14 @@ func (c *CustomFieldManager) GetString(name string, loc *Location) (string, erro
 	case *Number:
 		return string(*fv.(*Number)), nil
 	case SingleOption:
+		if string(fv.(SingleOption)) == "" {
+			return "", nil
+		}
 		return c.CustomFieldOptionName(name, string(fv.(SingleOption)))
 	case *SingleOption:
+		if string(*fv.(*SingleOption)) == "" {
```
mhupman, line 588: do we already handle the value being ==nil above?  If so, this is fine.  If not, it might blow up.

mhupman (approved):
lgtm

customfield_service.go
cdworak: Yes! handled above

PR #55, opened 2018-06-07
Add diff to list

cdworak:


list_diff.go
```diff
@@ -0,0 +1,19 @@
+package yext
+
+import "reflect"
+
+func (a *BioList) Diff(b *BioList) bool {
```
mhupman: Hmm, these don't really Diff, they look more like Equal() to me?

mhupman (approved):

PR #57, opened 2018-07-10
Additions to revert error code string to error object

angelaxu1:


error.go
```diff
@@ -101,3 +102,46 @@ func IsErrorCode(err error, code int) bool {
...
+
+func StrToError(str string) *Error {
+	strRemaining := strings.TrimLeft(str, "type: ")
+	typ, strRemaining := SplitStrAtWord(strRemaining, "code:")
+	code, message := SplitStrAtWord(strRemaining, "message:")
+	codeInt, _ := strconv.Atoi(code)
+	return &Error{Type: typ, Code: codeInt, Message: message}
+}
+
+func ParseErrorStr(errorStr string) []*Error {
```
cdworak: can we rename from ErrorsFromString(). It follows a similar model of our function naming seen in Hours

error.go
```diff
@@ -101,3 +102,46 @@ func IsErrorCode(err error, code int) bool {
...
+			if before != "" {
+				before += " "
+			}
+			before += w
+		}
+	}
+	return before, after
+}
+
+func StrToError(str string) *Error {
```
cdworak: and then rename this to ErrorFromString()

cdworak: you can also lowercase this function (strToError, or now errorFromString() because it's a helper function and probably shouldn't be needed outside this package)

error.go
cdworak: can we change the errorStr parameter name to str (to match the function above it)

error.go
```diff
@@ -101,3 +102,46 @@ func IsErrorCode(err error, code int) bool {
...
+	strRemaining := strings.TrimLeft(str, "type: ")
+	typ, strRemaining := SplitStrAtWord(strRemaining, "code:")
+	code, message := SplitStrAtWord(strRemaining, "message:")
+	codeInt, _ := strconv.Atoi(code)
```
cdworak: hmm we'll probably want to catch that error if the converstion doesn't go through ok (the strconv that is) and return the error

error.go
```diff
@@ -101,3 +102,46 @@ func IsErrorCode(err error, code int) bool {
 	}
 	return false
 }
+
+func SplitStrAtWord(str string, word string) (string, string) {
+	words := strings.Split(str, " ")
```
cdworak: can you put these in a var block? 

var (
words = strings.Split(
found = false
before, after = ""
)

error.go
```diff
@@ -101,3 +102,46 @@ func IsErrorCode(err error, code int) bool {
 	}
 	return false
 }
+
+func SplitStrAtWord(str string, word string) (string, string) {
```
cdworak: you can also lowercase this function because it is a helper (splitStringAtWord)

cdworak (commented):
Looks good! Could you add an error_test.go file and add some tests to see if your errors are parsing as expected? 

you can look at the other _test files in yext-go for example but I imagine your tests could look something like this:

test := []struct{
 ErrorMessage string
 Want []*Error
}{
 ErrorMessage: "This is the error message; This is another error message;",
 Want: []*Error{
 [These are the errors you want],

Let me know if you need help!

cdworak (approved):
This looks great! i can make an item for  you to add tests for next week (if you won't get to it today)

error_test.go
```diff
@@ -54,3 +56,54 @@ func TestIsNotFoundError(t *testing.T) {
...
 	}
 }
+
+func TestErrorsFromString(t *testing.T) {
+	tests := []struct {
+		ErrorMessage string
+		Want         []*Error
+	}{
+		{
+			ErrorMessage: "type: FATAL_ERROR code: 2015 message: Unknown folder; request uuid: 7199948d-9f0d-4649-9625-495b33ad4940",
```
cdworak: looks great! can you add one test for the scenario where the error message is empty? I assume it should return nil in that case?

PR #58, opened 2018-07-18
Entities

cdworak:
this isn't even really a pull request because it's not done, just so hup can see


client.go
```diff
@@ -324,7 +325,7 @@ func tokenListHelper(lr tokenListRetriever, opts *ListOptions) error {
 func addListOptions(requrl string, opts *ListOptions) (string, error) {
 	u, err := url.Parse(requrl)
 	if err != nil {
-		return "", nil
+		return "", err
```
mhupman, line 396: and a bugfix - hurrah!

client.go
```diff
@@ -336,6 +337,7 @@ func addListOptions(requrl string, opts *ListOptions) (string, error) {
 		q.Add("limit", strconv.Itoa(opts.Limit))
 	}
 	if opts.PageToken != "" {
+		// TODO: this is not compatible with the old parameter name
```
mhupman: :(

maybe the func could take another parameter if the new/old apis are the same in everything other than param name

customfield.go
```diff
@@ -231,18 +233,18 @@ func (v *VideoGallery) CustomFieldTag() string {
...
-	HolidayHours   []HolidayHours `json:"holidayHours,omitempty"`
-}
-
 func (h Hours) CustomFieldTag() string {
 	return CUSTOMFIELDTYPE_HOURS
 }
 
 type DailyTimes struct {
-	DailyTimes string `json:"dailyTimes,omitempty"`
+	Sunday    string `json:"sunday,omitempty"`
```
mhupman: was this wrong all along or a change in Entities?

customfield_service.go
```diff
@@ -531,6 +531,7 @@ func ParseCustomFields(cfraw map[string]interface{}, cfs []*CustomField) (map[st
 	return parsed, nil
 }
 
+// TODO: Can we do this anymore?
```
mhupman: probably not, and it probably doesn't make sense.  it was just to prevent the "obvious" mistake of passing in a string vs id.   now we probably won't know until we send to the API (unless we validate against the field schema, but that is probably a new exposed method, not an automagic thing)

customfield_service_test.go
```diff
@@ -42,11 +42,11 @@ func runParseTest(t *testing.T, index int, c customFieldParseTest) {
 	)
 
 	if cfType != expectType {
-		t.Errorf("test #%d (%s) failed: expected type %s, got type %s", index, c.TypeName, expectType, cfType)
+		t.Errorf("test #%d (%s) failed:\nexpected type %s\ngot type %s", index, c.TypeName, expectType, cfType)
```
mhupman: i really hate how hard it is to format things in test output

mhupman: the tabbing in particular

customfield_service_test.go
```diff
@@ -59,15 +59,58 @@ var (
...
 			map[string]interface{}{
-				"date":  "2016-05-30",
-				"hours": "",
+				"date": "2016-05-30",
 			},
 			map[string]interface{}{
-				"date":  "2016-05-31",
-				"hours": "9:00:17:00",
+				"date": "2016-05-31",
+				"hours": []*Times{
```
mhupman: you chose interface types above but a concrete type here - if you're trying to mimic json should be primitive values, like maps and strings?

mhupman: (the reason being that Times could have its own MarshalJSON define which would mess up your tests)

entity.go
```diff
@@ -0,0 +1,22 @@
+package yext
+
+type EntityType string
+
+// This is a little dangerous because if you embed an Entity (like Location) within another Entity
+// You don't have to re-define these...but we'd need to re-define copy...
```
mhupman: comment mentions "copy"

entity.go
```diff
@@ -0,0 +1,22 @@
...
+type Entity interface {
```
mhupman, line 9: love the narrow interface!

entity.go
```diff
@@ -0,0 +1,22 @@
...
+type EntityMeta struct {
+	Id          *string           `json:"id,omitempty"`
+	AccountId   *string           `json:"accountId,omitempty"`
+	EntityType  EntityType        `json:"entityType,omitempty"`
+	FolderId    *string           `json:"folderId,omitempty"`
+	LabelIds    *UnorderedStrings `json:"labelIds,omitempty"`
+	CategoryIds *[]string         `json:"categoryIds,omitempty"`
+	Language    *string           `json:"language,omitempty"`
+	CountryCode *string           `json:"countryCode,omitempty"`
+	nilIsEmpty  bool
```
mhupman: oh no do we still need this thing :(

entity.go
```diff
@@ -0,0 +1,22 @@
+package yext
```
mhupman, line 1: i might suggest adding an Entity type here that is appropriate as an Unmarshal target for any entity... basically just 

type BaseEntity struct {
 Meta *EntityMeta `json:"meta"`
}

maybe no other attributes.  it would also satisfy the Entity interface

entity_service.go
```diff
@@ -0,0 +1,253 @@
...
+		tmp = entity
+	}
+	newVal = reflect.New(reflect.TypeOf(tmp)).Elem().Interface()
+	e.registry[entityType] = newVal
+}
+
+func (e *EntityService) LookupEntity(entityType EntityType) (interface{}, error) {
+	val, ok := e.registry[entityType]
+	if !ok {
+		return nil, fmt.Errorf("Unable to find entity type %s in entity registry %v", entityType, e.registry)
```
mhupman: instead of spitting out the whole map, maybe just print out the names of the "known" types?

entity_service.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	return reflect.New(reflect.TypeOf(val)).Interface(), nil
+}
+
+func GetBytes(key interface{}) ([]byte, error) {
```
mhupman: heh too much copy/pasta

entity_service.go
```diff
@@ -0,0 +1,253 @@
...
+	if isPtr {
+		tmp = reflect.ValueOf(entity).Elem().Interface()
+	} else {
+		tmp = entity
+	}
+	newVal = reflect.New(reflect.TypeOf(tmp)).Elem().Interface()
+	e.registry[entityType] = newVal
+}
+
+func (e *EntityService) LookupEntity(entityType EntityType) (interface{}, error) {
```
mhupman: this doesn't just lookup but it also creates an instance of the type, right?  might adjust the name

entity_test.go
```diff
@@ -0,0 +1,356 @@
...
+	CFPhoto        *Photo      `json:"cf_Photo,omitempty"`
+	CFVideos       []*Video    `json:"cf_Videos,omitempty"`
+	CFVideo        *Video      `json:"cf_Video,omitempty"`
+	CFDate         *Date       `json:"cf_Date,omitempty"`
+	CFSingleOption *string     `json:"cf_SingleOtpion,omitempty"`
+	CFMultiOption  *[]string   `json:"cf_MultiOption,omitempty"`
+}
+
+func (l *CustomLocationEntity) EntityId() string {
+	return ""
```
mhupman: ?

entity_test.go
```diff
@@ -0,0 +1,356 @@
...
+}
+
+func (l *CustomLocationEntity) Type() EntityType {
+	return ENTITYTYPE_LOCATION
```
mhupman: ?

mhupman: can we get this info off of Meta, which is found in the embedded Location?

entity_test.go
```diff
@@ -0,0 +1,356 @@
...
+}
+
+func (l *CustomLocationEntity) PathName() string {
+	return locationsPath
+}
+
+func (l *CustomLocationEntity) Copy() Entity {
```
mhupman: do we need this?

error.go
```diff
@@ -101,3 +102,54 @@ func IsErrorCode(err error, code int) bool {
...
+	typ, strRemaining := splitStrAtWord(strRemaining, "code:")
+	code, message := splitStrAtWord(strRemaining, "message:")
+	codeInt, errConv := strconv.Atoi(code)
+	if errConv != nil {
+		return nil, errConv
+	}
+	return &Error{Type: typ, Code: codeInt, Message: message}, nil
+}
+
+func ErrorsFromString(errorStr string) ([]*Error, error) {
```
mhupman: why do we need to go backwards from the printed version of an error?

event.go
```diff
@@ -0,0 +1,34 @@
+package yext
```
mhupman: suggestion - prefix the filenames of these things so we know they go together - `entity_event.go`, for example

hydrate.go
```diff
@@ -2,7 +2,7 @@ package yext
 
 import "fmt"
 
-// TODO: This mutates the location, no need to return the value
+// TODO: I think this can be deleted (unless we keep the location-service around)
```
mhupman: FWIW i think we have to keep location service around as long as it remains in the API

language_profile_service.go
```diff
@@ -31,9 +31,9 @@ func (l *LanguageProfileService) GetAll(id string) (*LanguageProfileListResponse
 		return nil, r, err
 	}
 
-	if _, err := l.HydrateLocations(v.LanguageProfiles); err != nil {
-		return nil, r, err
-	}
+	// if _, err := l.HydrateLocations(v.LanguageProfiles); err != nil {
```
mhupman: commented out tests will need to be undone

location.go
```diff
@@ -10,13 +10,14 @@ import (
 	"fmt"
 )
 
+const ENTITYTYPE_LOCATION EntityType = "LOCATION"
```
mhupman: would suggest leaving this file alone and starting fresh with `entity_location.go` - location service needs to continue working as is

location_diff.go
```diff
@@ -28,7 +28,8 @@ type Comparable interface {
 //   // isDiff -> false
 //   // delta -> nil
 func (y Location) Diff(b *Location) (d *Location, diff bool) {
-	d = &Location{Id: String(y.GetId())}
+	d = &Location{EntityMeta: &EntityMeta{
```
mhupman: similar to above, would suggest leaving this alone and starting with an different differ for Entity usage.  i can't imagine these can share the same code

entity_service.go
```diff
@@ -0,0 +1,253 @@
...
+	ResolvePlaceholders bool
+}
+
+type EntityListResponse struct {
+	Count     int           `json:"count"`
+	Entities  []interface{} `json:"entities"`
+	PageToken string        `json:"pageToken"`
+}
+
+func (e *EntityService) RegisterDefaultEntities() {
```
mhupman, line 73: who calls this?  should probably make it automatic in the client bootstrapping, if possible.  maybe part of the Config() as a toggle?

mhupman: perhaps allowing it to be controlled isn't worth it

entity_service.go
```diff
@@ -0,0 +1,253 @@
...
+	if opts.ResolvePlaceholders {
+		q.Add("resolvePlaceholders", "true")
+	}
+	u.RawQuery = q.Encode()
+
+	return u.String(), nil
+}
+
+func (e *EntityService) Get(id string) (Entity, *Response, error) {
+	var v interface{}
```
mhupman: could make it a little easier on yourself and make this `map[string]interface{}`

entity_service.go
```diff
@@ -0,0 +1,253 @@
...
+	r, err := e.client.DoRequest("GET", fmt.Sprintf("%s/%s", entityPath, id), &v)
+	if err != nil {
+		return nil, r, err
+	}
+
+	entity, err := e.toEntityType(v)
```
mhupman: error handling?

entity_service.go
```diff
@@ -0,0 +1,253 @@
...
+		return nil, fmt.Errorf("Unable to find meta attribute in %v", entityValsByKey)
+	}
+
+	var metaByKey = meta.(map[string]interface{})
+	entityType, ok := metaByKey["entityType"]
+	if !ok {
+		return nil, fmt.Errorf("Unable to find entityType attribute in %v", metaByKey)
+	}
+
+	entityObj, err := e.LookupEntity(EntityType(entityType.(string)))
```
mhupman: we should think about what we want to happen if a type is NOT registered.  for example, what if the caller knew they didn't know the types ahead of time and wanted to just inspect them as JSON/map[string]interface{}?  we might want to let that through and not error out in some cases?

mhupman: on the other hand, i can also see it useful to know if you weren't able to unwrap the type, but maybe we find that our during type assertion?

entity_service.go
```diff
@@ -0,0 +1,253 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Convert into struct of Entity Type
+	entityJSON, err := json.Marshal(entityValsByKey)
+	if err != nil {
+		return nil, fmt.Errorf("Error marshaling entity to JSON: %s", err)
```
mhupman: don't include "error" in error message or capitalize first word

entity_service.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	err = json.Unmarshal(entityJSON, &entityObj)
+	if err != nil {
+		return nil, fmt.Errorf("Error unmarshaling entity JSON: %s", err)
```
mhupman: same

entity_service.go
```diff
@@ -0,0 +1,253 @@
...
+	var buf bytes.Buffer
+	enc := gob.NewEncoder(&buf)
+	err := enc.Encode(key)
+	if err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
+
+func (e *EntityService) toEntityTypes(entityInterfaces []interface{}) ([]Entity, error) {
```
mhupman: this is throughout the PR, but the param names are stuttering with their type... prefer shortened values or descriptions of the content without the type:

toEntityTypes(entities []interface{})

entity_service.go
```diff
@@ -0,0 +1,253 @@
...
+		entity, err := e.toEntityType(entityInterface)
+		if err != nil {
+			return nil, err
+		}
+		entities = append(entities, entity)
+	}
+	return entities, nil
+}
+
+func (e *EntityService) toEntityType(entityInterface interface{}) (Entity, error) {
```
mhupman: similarly, `toEntityType(entity interface{})` or `toEntityType(e interface{})`

mhupman (commented):
nice start!  i agree the Register/Unwrap might need some tweaking, but only usage will tell for sure

customfield_service_test.go
cdworak: yep will fix

entity.go
cdworak: I'm not sure yet but don't want to forget about it

entity_service.go
cdworak: Config is a little barebones, I'll make it automatic in the bootstrapping and we can always add it to the config later if we need to control it

entity_service.go
cdworak: yeah....i had a hard time because both the word type and interface you can't use in go, so i prefixed them a lot. I'll try to clean it up a bit

entity.go
cdworak: With the idea being that we would embed this guy in other Entities like Location? Or just have him around as the most stripped down entity?

entity_test.go
cdworak: I think i was worried that if someone creates a new Location type and doesn't set the type...then returning EntityType would give you an ""

But I suppose the same thing would happen if you don't set the Id so you would just need to not forget?

error.go
cdworak: I didn't write these....let me look into why they popped up all of a sudden

event.go
cdworak: I think this is some of the interns work for the summer that somehow got on this PR. i'll remove it

customfield_service.go
```diff
@@ -531,7 +531,6 @@ func ParseCustomFields(cfraw map[string]interface{}, cfs []*CustomField) (map[st
 	return parsed, nil
 }
 
-// TODO: Can we do this anymore?
 func validateCustomFieldsKeys(cfs map[string]interface{}) error {
```
mhupman: maybe we could rename this to validateLocationCustomFields or something to indicate it is legacy

registry.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"fmt"
+	"reflect"
+)
+
+type Registry map[string]interface{}
+
+func (r Registry) Register(key string, val interface{}) {
+	//From: https://github.com/reggo/reggo/blob/master/common/common.go#L169
+	isPtr := reflect.ValueOf(val).Kind() == reflect.Ptr
```
mhupman: var (..) for all these

entity.go
mhupman: yep, could embed to get Meta "for free" without need to duplicate.  it would also be a good deser target if you didn't know what you were going to get from the API - for example, if you hit the generic entities endpoint to return any and all entities (assuming that exists?)?

entity_test.go
mhupman: Oh, you mean on create?  If its pulled from the API, it would always be populated for you, right?

error.go
mhupman: whew!

event.go
mhupman: 😱

customfield.go
cdworak: a change in Entities...I copied over the old struct. same thing happened for hours. will figure out how to distinguish between the two

entity.go
cdworak: got it

entity_service.go
cdworak: added

cftasset_service.go
```diff
@@ -0,0 +1,148 @@
+package yext
```
mhupman, line 1: why is CFT Asset a different service?  can this be part of a unified AssetService?

cftasset.go
```diff
@@ -0,0 +1,152 @@
+package yext
```
mhupman, line 1: similarly, isn't this just asset.go?

asset.go
```diff
@@ -1,20 +1,11 @@
 package yext
```
mhupman, line 1: High level feedback discussed with @upalsaha:

* Maintain AssetService but rename it to LegacyAssetService
* Name the new service (with registry, entity and CFT support) AssetService
* Add a switch in NewClient() that decides what service to create based on the Version date - leave the other service `nil`

cc @cdworak

cftasset.go
```diff
@@ -0,0 +1,152 @@
...
+
+const (
+	ASSETTYPE_TEXT         AssetType = "text"
+	ASSETTYPE_IMAGE        AssetType = "image"
+	ASSETTYPE_VIDEO        AssetType = "video"
+	ASSETTYPE_COMPLEXIMAGE AssetType = "complexImage"
+	ASSETTYPE_COMPLEXVIDEO AssetType = "complexVideo"
+)
+
+type CFTAsset struct {
```
mhupman, line 40: per above, this would just be Asset - the old thing would become LegacyAsset

cftasset.go
```diff
@@ -0,0 +1,152 @@
...
+	LabelOperator LabelOperator     `json:"labelOperator,omitempty"`
+}
+
+type AssetUsage struct {
+	Type       UsageType         `json:"type,omitempty"`
+	FieldNames *UnorderedStrings `json:"fieldNames,omitempty"`
+}
+
+func (a *AssetUsage) Equal(b Comparable) bool {
+    defer func() {
```
mhupman: why is this necessary?

cftasset.go
```diff
@@ -0,0 +1,152 @@
...
+        if r := recover(); r != nil {
+            fmt.Printf("Value of A: %+v, Value of B:%+v, Type Of A: %T, Type Of B: %T\n", a, b, a, b)
+            panic(r)
+        }
+    }()
+
+    bAssetUsage := *b.(*AssetUsage)
```
mhupman: is this dereference safe?

cftasset.go
```diff
@@ -0,0 +1,152 @@
...
+        return false
+    }
+
+    return true
+}
+
+type AssetUsageList []AssetUsage
+
+func (a *AssetUsageList) Equal(b Comparable) bool {
+    defer func() {
```
mhupman: same

cftasset_diff.go
```diff
@@ -0,0 +1,21 @@
...
+	if a == nil && b != nil {
+		return b, true
+	} else if b == nil && a != nil {
+		return a, true
+	} else if a == nil && b == nil {
+		return nil, false
+	}
+
+	if a.GetAssetType() != b.GetAssetType() {
+		return nil, true
```
mhupman, line 13: nil?

cftasset_diff.go
```diff
@@ -0,0 +1,21 @@
+package yext
+
+func (a *CFTAsset) Diff(b *CFTAsset) (*CFTAsset, bool) {
+	if a == nil && b != nil {
+		return b, true
+	} else if b == nil && a != nil {
+		return a, true
```
mhupman, line 7: if b is nil there are no proposed changes, right?  wouldn't that be `nil, false`

cftasset_service.go
```diff
@@ -0,0 +1,148 @@
...
+func (a *CFTAssetService) RegisterDefaultAssetValues() {
+	a.registry = make(Registry)
+	// ASSETTYPE_TEXT doesn't need to register because it's just a string
+	a.RegisterAssetValue(ASSETTYPE_IMAGE, &ImageValue{})
+	a.RegisterAssetValue(ASSETTYPE_VIDEO, &VideoValue{})
+	a.RegisterAssetValue(ASSETTYPE_COMPLEXIMAGE, &ComplexImageValue{})
+	a.RegisterAssetValue(ASSETTYPE_COMPLEXVIDEO, &ComplexVideoValue{})
+}
+
+func (a *CFTAssetService) RegisterAssetValue(t AssetType, assetValue interface{}) {
```
mhupman, line 31: just curious - any reason we have to wrap these functions vs exposing the Registry directly?

cftasset_service.go
```diff
@@ -0,0 +1,148 @@
...
+	for _, asset := range assets {
+		if err := a.toAssetWithValue(asset); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (a *CFTAssetService) toAssetWithValue(asset *CFTAsset) error {
+	if asset.Type == ASSETTYPE_TEXT {
```
mhupman, line 49: i think i asked this before, but why not just register ASSETTYPE_TEXT so you don't have to special case this?

customfield.go
```diff
@@ -231,20 +233,34 @@ func (v *VideoGallery) CustomFieldTag() string {
 	return CUSTOMFIELDTYPE_VIDEO
 }
 
-type Hours struct {
-	AdditionalText string         `json:"additionalHoursText,omitempty"`
-	Hours          string         `json:"hours,omitempty"`
-	HolidayHours   []HolidayHours `json:"holidayHours,omitempty"`
+// HoursCustom is the Hours custom field format used by locations API
+// Entities API uses the Hours struct in location_entities.go (profile and custom hours are defined the same way for entities)
+type HoursCustom struct {
```
mhupman: if we follow the other pattern, maybe LegacyHours?

entity.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (b *BaseEntity) GetNilIsEmpty() bool {
+	return b.nilIsEmpty
+}
+
+func (b *BaseEntity) SetNilIsEmpty(val bool) {
+	b.nilIsEmpty = val
+}
+
+type RawEntity map[string]interface{}
```
mhupman, line 132: another option is to do:

struct RawEntity {
  Meta Meta
  fields: map[string]interface
}

and implement json.Unmarshaler on it.  then you wouldn't need to parse meta by hand

mhupman: yeah i like this

event.go
mhupman: bump on removal

location_diff.go
```diff
@@ -148,7 +150,7 @@ func isZeroValue(v reflect.Value, interpretNilAsZeroValue bool) bool {
 		return isZeroValue(v.Elem(), interpretNilAsZeroValue)
 	case reflect.Struct:
 		for i, n := 0, v.NumField(); i < n; i++ {
-			if !isZeroValue(v.Field(i), interpretNilAsZeroValue) {
+			if !isZeroValue(v.Field(i), true) {
```
mhupman: is this a change we want in location_diff

mhupman: (vs entity_diff)

location_entity.go
```diff
@@ -0,0 +1,670 @@
...
+func (y LocationEntity) GetLabelIds() (v UnorderedStrings) {
+	if y.LabelIds != nil {
+		v = *y.LabelIds
+	}
+	return v
+}
+
+func (y LocationEntity) SetLabelIds(v []string) {
+	l := UnorderedStrings(v)
+	y.SetLabelIdsWithUnorderedStrings(l)
```
mhupman: can you inline the UnorderedStrings wrapper?

location_entity.go
```diff
@@ -509,20 +506,20 @@ func (y LocationEntity) GetCategoryIds() (v []string) {
 	return v
 }
 
-func (y LocationEntity) GetLabelIds() (v UnorderedStrings) {
-	if y.LabelIds != nil {
-		v = *y.LabelIds
+func (y LocationEntity) GetLabels() (v UnorderedStrings) {
+	if y.BaseEntity.Meta.Labels != nil {
```
cdworak: we should probably also move these functions under BaseEntity

PR #59, opened 2018-08-21
Implement Diff() functions for Roles, ACLs, and ACLLists and add tests

bharatRao:


acl_diff.go
```diff
@@ -0,0 +1,50 @@
+package yext
+
+// Diff calculates the differences between a base ACL (a) and a second ACL (b).
+// The result returned is an ACL object with only parameters that are strictly different.
+func (a ACL) Diff(b ACL) (delta ACL, diff bool) {
+	roleDelta, roleDiff := a.Role.Diff(b.Role)
+	if !roleDiff && a.On == b.On && a.AccessOn == b.AccessOn {
+		return ACL{}, false
```
benmcginnis: this should probably actually return a pointer to an ACL, so you can return nil

acl_diff.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Diff for ACLList calculates if ACLList a is identical to ACLList b (order is ignored).
+// If any ACL in a is missing in b, return the entire list b.
+func (a ACLList) Diff(b ACLList) (delta ACLList, diff bool) {
+	if len(a) != len(b) {
+		return b, true
+	}
+
+	for _, aclA := range a {
+		for idxB, aclB := range b {
```
benmcginnis: I think this style is easier to understand:
```
for _, aclA := range a {
  var found bool
  for _, aclB := range b {
    if _, hasDelta := aclA.Diff(aclB); !hasDelta {
       found = true
    }
  }  
 if !found {
     return b, true
  }
}
```

benmcginnis: Also, I think we need to do the reverse, right?

a test case that would probably fail would be `([]ACL{a, a, c}).Diff([]ACL{a,b,c})` right, because it would return `nil, false`

acl_diff_test.go
```diff
@@ -0,0 +1,186 @@
...
+			Name: differentRoleACL.Name,
+		},
+	}
+
+	if d != expectedDiffACL {
+		t.Errorf("Expected %v, but got %v", expectedDiffACL, d)
+	}
+}
+
+func TestDiffOnACL(t *testing.T) {
```
benmcginnis: The more canonical way to structure these test functions is:

```
func TestACL_Diff(t *testing.T) {
	tests := []struct {
		name      string
		haveA     ACL
		haveB     ACL
		wantDelta ACL
		wantDiff  bool
	}{
		{
			"Identical",
			ACL{
				Role: yext.Role{
					Id:   yext.String("3"),
					Name: yext.String("Example Role"),
				},
				On:       "12345",
				AccessOn: yext.ACCESS_FOLDER,
			},

			ACL{
				Role: yext.Role{
					Id:   yext.String("3"),
					Name: yext.String("Example Role"),
				},
				On:       "12345",
				AccessOn: yext.ACCESS_FOLDER,
			},
			nil,
			false,
		},
	}

	for _, test := range tests {
		if gotDelta, gotDiff := test.haveA.Diff(test.haveB); wantDiff != gotDiff {
			t.Error(fmt.Sprintf("test '%s' failed, got diff: %t wanted %t, wanted delta: %+v got delta: %+v", test.name, test.wantDiff, gotDiff, test.wantDelta, gotDelta))
		}
	}
}
```

acl_diff_test.go
```diff
@@ -0,0 +1,186 @@
...
+	if isDiff {
+		t.Errorf("Expected diff to be false but was true, diff result: %v", d)
+	}
+
+	if d != nil {
+		t.Errorf("Expected %v, but got %v", nil, d)
+	}
+}
+
+func TestDiffItemACLList(t *testing.T) {
```
benmcginnis: then you would have a separate test function for `TestACLList_Diff` that followed the same pattern as above

role_diff_test.go
```diff
@@ -0,0 +1,72 @@
...
+		Name: yext.String("Example Role"),
+	}
+
+	differentNameRole = yext.Role{
+		Id:   yext.String("3"),
+		Name: yext.String("Example Role Two"),
+	}
+)
+
+func TestDiffIdenticalRole(t *testing.T) {
```
benmcginnis: and here you would have `TestRole_Diff`

benmcginnis (changes requested):
almost there!

acl_diff.go
bharatRao: I guess that's true, but there should never be two identical ACLs anyway, right..?

bharatRao: Also, you meant: if found return b, true, not !found.

Edit: Sorry, actually the hasDelta check before doing found = true needs to be !hasDelta.

benmcginnis: actually what i meant was if !hasDelta; found = true

benmcginnis: and regarding the backwards direction, I think it's important that methods in libraries like this are robust, so I think we should handle the case.

benmcginnis (approved):
lgtm!

benmcginnis:
please squash before merging

upalsaha (approved):
LGTM

bharatRao:
I merged it, just not via the multipleDiffMethods branch.

PR #62, opened 2018-10-03
CFT Assets: modify asset data model

upalsaha:
What I've got so far from reading through the spec and playing with the new endpoints. Just wanted some eyes on this and spoke with @mhupman about needing a type registry. Was able to create a complexImage with what's here

### DIFF
--- a/asset.go
+++ b/asset.go
@@ -1,11 +1,21 @@
 package yext
 
-type AssetType string
-
 const (
-	AssetTypeText  AssetType = "AssetTypeText"
-	AssetTypePhoto AssetType = "AssetTypePhoto"
-	AssetTypeVideo AssetType = "AssetTypeVideo"
+	AssetTypeText         = "text"
+	AssetTypeImage        = "image"
+	AssetTypeVideo        = "video"
+	AssetTypeComplexImage = "complexImage"
+	AssetTypeComplexVideo = "complexVideo"
+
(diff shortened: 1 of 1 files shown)


asset.go
```diff
@@ -15,30 +25,44 @@ const (
...
+	Id          *string       `json:"id"`
+	Name        *string       `json:"name"`
+	Type        *string       `json:"type"`
+	Description *string       `json:"description,omitempty"`
+	ForEntities *ForEntities  `json:"forEntities,omitempty"`
+	Usage       *[]AssetUsage `json:"usage,omitempty"`
+	Locale      *string       `json:"locale,omitempty"`
+	Labels      *[]string     `json:"labels,omitempty"`
+	Owner       *uint64       `json:"owner,omitempty"`
+	Value       *AssetValue   `json:"value,omitempty"`
```
mhupman, line 40: is this _only_ for the case of image/video Assets?  Are CFTs found under another key?

upalsaha: No, this "value" key will be any kind of value including the basic text/image/video types as well as CFT values. The type itself is denoted by the "type" key. Initially had this as an interface in the struct, but used that AssetValue type as a wrapper for the basic value types in my tests today

mhupman: are you sure that will work if the "value" isn't nested under another object?  it might have to be an interface{} type

PR #63, opened 2018-10-05
CFT Assets: modify asset data model, add asset type registry to asset service

upalsaha:


asset.go
```diff
@@ -15,30 +25,91 @@ const (
...
+	Image           *Image `json:"image,omitempty"`
+	ClickthroughURL string `json:"clickthroughUrl,omitempty"`
+	Details         string `json:"details,omitempty"`
+}
+
+func (ci ComplexImageValue) GetAssetType() AssetType {
+	return ASSETTYPE_COMPLEXIMAGE
+}
+
+type VideoValue struct {
```
mhupman: What are the wrapper types for?  I thought the value was directly under the “value” key

upalsaha: Not necessarily for the regular `image` and `video` types. For example, the response for a `image` asset looks like this:
```
"response": {
        "id": "upal2",
        "type": "image",
        "name": "Upal's image 2",
        "description": "testing image create",
        "value": {
            "image": {
                "url": "https://blah.com/image.jpg"
            }
        },
        "forEntities": {
            "mappingType": "ALL_ENTITIES"
        },
        "usage": [
            {
                "type": "ALL_PROFILE_FIELDS"
            }
        ],
        "owner": 012345
    }
```

mhupman: Really?  That seems wrong.  Is that covered in the docs anywhere?  I checked https://docs.google.com/document/d/1CinIC-UwTCIWqmBwavu8v1S2p_CYtGXokUzhH3CHg2U/edit# but don't see it.

upalsaha: @mhupman Fixed in latest commit

asset.go
```diff
@@ -15,30 +25,91 @@ const (
...
+
+const (
+	ASSETTYPE_TEXT         AssetType = "text"
+	ASSETTYPE_IMAGE        AssetType = "image"
+	ASSETTYPE_VIDEO        AssetType = "video"
+	ASSETTYPE_COMPLEXIMAGE AssetType = "complexImage"
+	ASSETTYPE_COMPLEXVIDEO AssetType = "complexVideo"
+)
+
+type AssetUsageType string
```
cdworak: are you still using this?

asset_service.go
```diff
@@ -1,27 +1,85 @@
...
+	for _, asset := range assets {
+		if err := a.toAssetWithValue(asset); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (a *AssetService) toAssetWithValue(asset *Asset) error {
+	if asset.Type == ASSETTYPE_TEXT {
```
cdworak: I'm a bit confused how this will work for Text if you're returning nil here?

asset.go
```diff
@@ -15,30 +25,91 @@ const (
...
+	LabelIds      *UnorderedStrings `json:"labelIds,omitempty"`
+	LabelOperator *LabelOperator    `json:"labelOperator,omitempty"`
+}
+
+type AssetUsage struct {
+	Type       UsageType `json:"type"`
+	FieldNames *[]string `json:"fieldNames,omitempty"`
+}
+
+type AssetValue interface {
```
cdworak: are we using this interface anywhere?

asset_service.go
upalsaha: This works because the value of `value` is already just a plain string and doesn't need to be transformed, but in my newest commit I'm asserting this as a string

cdworak (approved):

asset.go
```diff
@@ -15,30 +25,63 @@ const (
...
+}
+
+type AssetUsage struct {
+	Type       UsageType `json:"type,omitempty"`
+	FieldNames *[]string `json:"fieldNames,omitempty"`
+}
+
+type TextValue string
+
+type ImageValue struct {
```
mhupman: thoughts on the "Value" suffix?  Every struct is a value, in a way....

asset_service.go
mhupman: Do you need to have this block if you registered "string" as a type on #24?

mhupman (approved):

entity.go
```diff
@@ -8,14 +8,13 @@ type Entity interface {
...
 type EntityMeta struct {
-	Id          *string           `json:"id,omitempty"`
-	AccountId   *string           `json:"accountId,omitempty"`
-	EntityType  EntityType        `json:"entityType,omitempty"`
-	FolderId    *string           `json:"folderId,omitempty"`
-	LabelIds    *UnorderedStrings `json:"labelIds,omitempty"`
-	CategoryIds *[]string         `json:"categoryIds,omitempty"`
-	Language    *string           `json:"language,omitempty"`
-	CountryCode *string           `json:"countryCode,omitempty"`
+	Id          *string    `json:"id,omitempty"`
```
cdworak: this does not match current entities branch

cdworak (approved):

PR #67, opened 2018-11-28
Id swap

cdworak:


location_diff_test.go
```diff
@@ -1244,6 +1244,20 @@ func TestDiffIds(t *testing.T) {
...
+			BaseLoc: &Location{
+				Id: String("1"),
+			},
+			NewLoc: &Location{
+				Id:   String("1"),
+				Name: String("New Name"),
+			},
+			IsDiff: true,
+			Delta: &Location{
+				Id:   String("1"), // Historically, even though there is no diff in the Id, we always want the Id set
```
cdworak, line 1257: @mhupman do you think setting the id even if it's the same is still something we want to continue now? It's how we've done it historically so I'm afraid to change it but wondering if it's more confusing now?

mhupman: great question!  I like the idea of not special casing ID, so perhaps we just require that the the caller figure it out.

mhupman (approved):

PR #68, opened 2018-12-05
J=PC-37965 add bulk endpoints to license service for v1

byronharvey24:
Here is a gist (dev account) with the request body as locationIds https://gist.github.com/cdworak/6c71dad906ce45540023d2635b2eb713

from postman both puts and deletes work with this format. (I cut it down to 3 locations for simplicity).
https://www.yext.com/s/703630/subscriptions/licenseHistory?locationId=0B3192

### DIFF
--- a/license.go
+++ b/license.go
@@ -1,9 +1,9 @@
 package yext
 
 type LicensePack struct {
-	LocationIds []string `json:"locationIds"`
-	Features    []string `json:"features"`
-	Id          int      `json:"id"`
-	Quantity    int      `json:"quantity"`
-	Status      string   `json:"status"`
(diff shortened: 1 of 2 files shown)


kunal-p (approved):

cdworak (approved):

cdworak (approved):

mhupman:
@byronharvey24 A few things for the future:
* Please squash your changes prior to merge if they don't need to be separate commits
* Watch the commit messages - "update to make a JSON request" doesn't provide any context about what was changed and why
* This is a public repository so we don't need to reference the relevant JIRA issue (any internal commits that use these changes should, of course)

PR #69, opened 2018-12-17
Add event entity support R=cdworak,usaha,bharvey

trob-indie:


event.go
```diff
@@ -8,10 +8,400 @@ const ENTITYTYPE_EVENT EntityType = "event"
 
 type EventEntity struct {
 	BaseEntity
-	Id          *string    `json:"id,omitempty"`
-	Name        *string    `json:"name,omitempty"`
-	Description *string    `json:"description,omitempty"`
-	EntityType  EntityType `json:"entityType,omitempty"`
+
+	// Admin
+	CategoryIds *[]string `json:"categoryIds,omitempty"`
```
upalsaha: should probably be UnorderedStrings, also isn't this in EntityMeta? I was having issues with other meta information not getting unmarshalled properly because of how I nested these fields

location_entity.go
```diff
@@ -90,9 +87,9 @@ type LocationEntity struct {
 	UberTripBrandingUrl  *string `json:"uberTripBrandingUrl,omitempty"`
 
 	// Social Media
-	FacebookCoverPhoto     *LocationPhoto `json:"facebookCoverPhoto,omitempty"`
-	FacebookPageUrl        *string        `json:"facebookPageUrl,omitempty"`
-	FacebookProfilePicture *LocationPhoto `json:"facebookProfilePicture,omitempty"`
+	FacebookCoverPhoto   *LocationPhoto `json:"facebookCoverPhoto,omitempty"`
+	FacebookPageUrl      *string        `json:"facebookPageUrl,omitempty"`
+	FacebookProfilePhoto *LocationPhoto `json:"facebookProfilePhoto,omitempty"`
```
upalsaha, line 94: someone I knew in high school made fun of me for saying Photo instead of Picture... look who has the last laugh

location_entity.go
```diff
@@ -119,6 +116,7 @@ type Address struct {
 	Region      *string `json:"region,omitempty"`
 	Sublocality *string `json:"sublocality,omitempty"`
 	PostalCode  *string `json:"postalCode,omitempty"`
+	CountryCode *string `json:"countryCode,omitempty"`
```
upalsaha: I hate that this is in two places

upalsaha (commented):
one real suggestion

event.go
```diff
@@ -8,10 +8,400 @@ const ENTITYTYPE_EVENT EntityType = "event"
...
+		return *y.Hours.HolidayHours
+	}
+	return nil
+}
+
+func (y EventEntity) GetBrands() (v []string) {
+	if y.Brands != nil {
+		v = *y.Brands
+	}
+	return v
```
byronharvey24, line 229: nil?

event.go
```diff
@@ -8,10 +8,400 @@ const ENTITYTYPE_EVENT EntityType = "event"
...
+		return *y.GoogleAttributes
+	}
+	return nil
+}
+
+func (y EventEntity) GetPaymentOptions() (v []string) {
+	if y.PaymentOptions != nil {
+		v = *y.PaymentOptions
+	}
+	return v
```
byronharvey24, line 355: same q here

event.go
```diff
@@ -8,10 +8,400 @@ const ENTITYTYPE_EVENT EntityType = "event"
...
+		return *y.TicketAvailability
+	}
+	return ""
+}
+
+func (b *EventEntity) GetCategoryIds() (v []string) {
+	if b.CategoryIds != nil {
+		v = *b.CategoryIds
+	}
+	return v
```
byronharvey24, line 411: one more time

event.go
upalsaha: @byronharvey24 These will be nil if it passes the "if" check above anyway

event.go
trob-indie: v is nil if it doesn't make it in the if statement. So this is the same thing.

event.go
trob-indie: It's not in EntityMeta according to https://developer.yext.com/docs/api-reference/#operation/createEntity

event.go
```diff
@@ -8,10 +8,407 @@ const ENTITYTYPE_EVENT EntityType = "event"
...
+	End   *string `json:"end,omitempty"`
+}
+
+type TicketPriceRange struct {
+	MinValue     *string `json:"minValue,omitempty"`
+	MaxValue     *string `json:"maxValue,omitempty"`
+	CurrencyCode *string `json:"currencyCode,omitempty"`
+}
+
+type PhotoGalleryItem struct {
```
cdworak, line 92: I think this is the same as Photo in LocationEntity? if so can you use that? Image should also be in LocationEntity

cdworak: Oh I don't think I've pushed that yet. Ok note to self to consolidate when pushed.

cdworak (approved):

PR #70, opened 2018-12-18
Diff Tests: Add CFTAsset diff() tests, fix circular import issue brea…

upalsaha:
…king acl and role diff tests


cdworak (commented):
this looks like it was included in a different push

PR #74, opened 2019-01-18
Hard-encode all requests to the yext api for # characters

byronharvey24:
The reason we can't use the url package is that the # character is a reserved character for URLs. There are others that can be viewed here https://www.lifewire.com/encoding-urls-3467463. That is the reason for hardcoding the encoded value for the # character. Going forward if we don't like using this hack it might make the most sense to tell clients not to use reserved url characters in their ids.

### DIFF
--- a/client.go
+++ b/client.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


upalsaha (approved):
approving for now but should probably look into longer term solution

PR #77, opened 2019-01-29
Updated entity code to accomodate varying Google Attributes fields

trob-indie:


event.go
```diff
@@ -70,7 +70,7 @@ type EventEntity struct {
 	PhotoGallery *[]PhotoGalleryItem `json:"photoGallery,omitempty"`
 	Videos       *[]Video            `json:"videos,omitempty"`
 
-	GoogleAttributes *GoogleAttributes `json:"googleAttributes,omitempty"`
+	GoogleAttributes *map[string]interface{} `json:"googleAttributes,omitempty"`
```
upalsaha: wouldn't it need to be a map of bools instead of interface types?

trob-indie: Json looks like this:

`                "googleAttributes": {
                    "pay_debit_card": [
                        "true"
                    ],
                    "requires_cash_only": [
                        "false"
                    ],
                    "has_gift_wrapping": [
                        "false"
                    ],
                    "has_delivery_same_day": [
                        "false"
                    ]
                },`

trob-indie: So it's actually a map of strings to a list of strings

upalsaha: oof, might want to like change to that then. Should avoid interface types if we know the expected types

byronharvey24: yeah wouldn't it be more accurate to do map[string][]string?

upalsaha (approved):

event.go
trob-indie: yep I updated

PR #78, opened 2019-02-07
Entities etl rebase

cdworak:


customfield_service.go
```diff
@@ -2,17 +2,53 @@ package yext
...
+		if err != nil {
+			return 0, 0, err
+		}
+		customFields = append(customFields, cfr.CustomFields...)
+		return len(cfr.CustomFields), cfr.Count, err
+	}
+
+	if err := listHelper(lr, &ListOptions{Limit: CustomFieldListMaxLimit}); err != nil {
+		return nil, err
+	} else {
```
byronharvey24: don't need this else statement, can just keep the syntactic go and return below the if

customfield_service.go
```diff
@@ -2,17 +2,53 @@ package yext
...
+}
+
+type CustomFieldResponse struct {
+	Count        int            `json:"count"`
+	CustomFields []*CustomField `json:"customFields"`
+}
+
+func (s *CustomFieldService) ListAll() ([]*CustomField, error) {
+	var customFields []*CustomField
+	var lr listRetriever = func(opts *ListOptions) (int, int, error) {
```
byronharvey24, line 24: is there any use in either splitting out this as a non-anonymous function or naming the returns since they are the same type?

customfield_service.go
```diff
@@ -2,17 +2,53 @@ package yext
...
+		cfr, _, err := s.List(opts)
```
byronharvey24, line 25: you don't want to check the status code of the response? Couldn't you get a 4xx and then cfr would be nil?

upalsaha: @byronharvey24 I think the thought here is that err will have the error body unmarshalled, which is definitely more verbose than just a status code

customfield_service.go
upalsaha: +2

customfield_service.go
```diff
@@ -304,315 +167,47 @@ func (c *CustomFieldManager) MustCustomFieldOptionId(fieldName, optionName strin
...
-	return s.CustomFieldManager.CustomFields, nil
+func (c *CustomFieldManager) NullSingleOption() **string {
+	return NullString()
 }
 
-func (s *CustomFieldService) MustCacheCustomFields() []*CustomField {
-	slice, err := s.CacheCustomFields()
-	if err != nil {
-		panic(err)
+func (c *CustomFieldManager) MustMultiOptionIds(fieldName string, optionNames ...string) *[]string {
```
byronharvey24, line 184: is there a reason we are returning a pointer to the slice instead of just the slice?

customfield_service.go
```diff
@@ -304,315 +167,47 @@ func (c *CustomFieldManager) MustCustomFieldOptionId(fieldName, optionName strin
...
-	return parsed, nil
-}
-
-// validateLocationCustomFieldsKeys can be used with Location API to validate custom fields
-func validateLocationCustomFieldsKeys(cfs map[string]interface{}) error {
-	for k, _ := range cfs {
-		if !customFieldKeyRegex.MatchString(k) {
-			return errors.New(fmt.Sprintf("custom fields must be specified by their id, not name: %s", k))
+	for _, v := range *setOptionIds {
+		if v == optionIds[0] {
```
byronharvey24: why the 0th?

entity_diff.go
```diff
@@ -10,12 +10,24 @@ func instanceOf(val interface{}) interface{} {
...
 		tmp   interface{}
 	)
+
 	if isPtr {
-		tmp = reflect.ValueOf(val).Elem().Interface()
-	} else {
-		tmp = val
+		var ptr = reflect.New(reflect.TypeOf(val).Elem()).Interface()
+		var numPointers = 0
+		for reflect.ValueOf(val).Kind() == reflect.Ptr {
```
upalsaha, line 19: nice

entity_diff.go
```diff
@@ -10,12 +10,24 @@ func instanceOf(val interface{}) interface{} {
 		isPtr = reflect.ValueOf(val).Kind() == reflect.Ptr
 		tmp   interface{}
 	)
+
 	if isPtr {
-		tmp = reflect.ValueOf(val).Elem().Interface()
-	} else {
-		tmp = val
+		var ptr = reflect.New(reflect.TypeOf(val).Elem()).Interface()
+		var numPointers = 0
```
byronharvey24: these two could be var blocked

entity_service.go
```diff
@@ -13,13 +14,20 @@ const (
 
 type EntityService struct {
 	client   *Client
-	registry Registry
+	Registry *EntityRegistry
```
upalsaha, line 17: do we need this exported? think it would be safer if we manipulated the registry with EntityService receiver functions rather than allowing it to be changed directly

entity_diff.go
```diff
@@ -10,12 +10,24 @@ func instanceOf(val interface{}) interface{} {
...
+		for reflect.ValueOf(val).Kind() == reflect.Ptr {
+			val = reflect.ValueOf(val).Elem().Interface()
```
byronharvey24, line 20: So coming out of this function val will be the last time that the equality was reached. Is that desirable? Are we assuming just one occurrence?

entity_service.go
```diff
@@ -13,13 +14,20 @@ const (
...
 type EntityListOptions struct {
 	ListOptions
 	SearchID            string
 	ResolvePlaceholders bool
+	EntityTypes         []string
+}
+
+// Used for Create and Edit
+type EntityServiceOptions struct {
+	TemplateId     string   `json:"templateId,omitempty"`
```
upalsaha, line 29: woot woot

entity_diff.go
```diff
@@ -87,12 +99,12 @@ func diff(a interface{}, b interface{}, nilIsEmptyA bool, nilIsEmptyB bool) (int
 			// Handle case where new is &Address{} and base is &Address{"Line1"}
 			if isZeroValue(valB, nilIsEmptyB) && !isZeroValue(valA, nilIsEmptyA) {
```
byronharvey24, line 102: do we not want to handle the reverse case if isZeroValue(valA) && !isZeroValue(valB)?

entity_diff_test.go
```diff
@@ -162,214 +162,214 @@ func TestEntityDiff(t *testing.T) {
...
-			deltaValue: Float(0),
+			deltaValue: NullableFloat(0),
 		},
 		diffTest{
-			name:      "*Float: both are 0 (E)",
+			name:      "**Float: both are 0 (E)",
 			property:  "YearEstablished",
-			baseValue: Float(2018),
+			baseValue: NullableFloat(2018),
 			newValue:  nil,
```
byronharvey24, line 201: why do we want this to be false?

entity_diff_test.go
```diff
@@ -162,214 +162,214 @@ func TestEntityDiff(t *testing.T) {
...
 			baseValue: nil,
 			newValue:  nil,
 			isDiff:    false,
 		},
 		diffTest{
-			name:      "*Float: base is not 0, new is nil (G)",
+			name:      "**Float: base is not 0, new is nil (G)",
 			property:  "YearEstablished",
-			baseValue: Float(1993),
+			baseValue: NullableFloat(1993),
```
byronharvey24, line 214: same here

entity_diff_test.go
```diff
@@ -162,214 +162,214 @@ func TestEntityDiff(t *testing.T) {
...
 			isDiff:     true,
-			deltaValue: Float(0),
+			deltaValue: NullableFloat(0),
 		},
 		diffTest{
-			name:      "*Float: base is 0, new is nil (J)",
+			name:      "**Float: base is 0, new is nil (J)",
 			property:  "YearEstablished",
-			baseValue: Float(0),
+			baseValue: NullableFloat(0),
```
byronharvey24, line 237: Why do we want this to be false and the one above this (I) to be true?

entity_diff_test.go
```diff
@@ -162,214 +162,214 @@ func TestEntityDiff(t *testing.T) {
...
 			newValue:  nil,
 			isDiff:    false,
 		},
 		diffTest{
-			name:      "*Bool: base is non-zero, new is nil (G)",
-			property:  "SuppressAddress",
-			baseValue: Bool(true),
+			name:      "**Bool: base is non-zero, new is nil (G)",
+			property:  "Closed",
+			baseValue: NullableBool(true),
```
byronharvey24, line 317: why is this not a diff?

entity_service.go
```diff
@@ -112,6 +120,28 @@ func (e *EntityService) List(opts *EntityListOptions) (*EntityListResponse, *Res
 	return v, r, nil
 }
 
+func addEntityServiceOptions(requrl string, opts *EntityServiceOptions) (string, error) {
+	u, err := url.Parse(requrl)
+	if err != nil {
+		return "", err
+	}
+
+	if opts == nil {
```
byronharvey24, line 129: Do we not want to return u here?

entity_service.go
```diff
@@ -187,7 +220,52 @@ func (e *EntityService) Create(y Entity) (*Response, error) {
 	return r, nil
 }
 
-func (e *EntityService) EditWithId(id string, y Entity) (*Response, error) {
+func (e *EntityService) CreateWithOptions(y Entity, opts *EntityServiceOptions) (*Response, error) {
+	var (
+		requrl = entityPath
```
byronharvey24, line 225: entityPath? Is that some global thing that I am missing here?

entity_service.go
```diff
@@ -187,7 +220,52 @@ func (e *EntityService) Create(y Entity) (*Response, error) {
...
+
+	u, err := url.Parse(requrl)
+	if err != nil {
+		return nil, err
+	}
+	q := u.Query()
+	q.Add("entityType", string(y.GetEntityType()))
+	u.RawQuery = q.Encode()
+	r, err := e.client.DoRequestJSON("POST", u.String(), y, nil)
+	if err != nil {
```
byronharvey24: instead of doing the if check you can just return r, err

entity_service.go
```diff
@@ -187,7 +220,52 @@ func (e *EntityService) Create(y Entity) (*Response, error) {
...
+		err    error
+	)
+
+	requrl, err = addEntityServiceOptions(requrl, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	r, err := e.client.DoRequestJSON("PUT", requrl, y, nil)
+	if err != nil {
```
byronharvey24, line 256: can just return r, err here as well

entity_service.go
```diff
@@ -187,7 +220,52 @@ func (e *EntityService) Create(y Entity) (*Response, error) {
...
+		return r, err
+	}
+
+	return r, nil
+}
+
+func (e *EntityService) EditWithOptions(y Entity, id string, opts *EntityServiceOptions) (*Response, error) {
+	var (
+		requrl = fmt.Sprintf("%s/%s", entityPath, id)
```
byronharvey24, line 246: this magical entityPath strikes again!

entity_service_test.go
```diff
@@ -45,3 +48,97 @@ func TestSetNilIsEmpty(t *testing.T) {
...
+				t.Errorf("Wanted token %s, got %s", test.token, v)
+			}
+			if v := r.URL.Query().Get("searchId"); v != test.searchID {
+				t.Errorf("Wanted searchId %s, got %s", test.searchID, v)
+			}
+			if v := r.URL.Query().Get("entityTypes"); v != test.entityTypes {
+				t.Errorf("Wanted entityTypes %s, got %s", test.entityTypes, v)
+			}
+			v := r.URL.Query().Get("resolvePlaceholders")
+			if v == "true" && !test.resolvePlaceholders || v == "" && test.resolvePlaceholders || v == "false" && test.resolvePlaceholders {
```
byronharvey24, line 136: I think this would be clearer if we wrapped the sections with parentheses

entity_test.go
```diff
@@ -6,19 +6,52 @@ import (
...
+		return err
+	}
+	if err := json.Unmarshal(bytes, &y.CustomEntity); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (y CustomLocationEntity) String() string {
+	b, _ := json.Marshal(y)
```
byronharvey24, line 40: Should check err here

entity_test.go
```diff
@@ -6,19 +6,52 @@ import (
...
+}
+
+func (c *CustomEntity) UnmarshalJSON(data []byte) error {
+	type Alias CustomEntity
+	a := &struct {
+		*Alias
+	}{
+		Alias: (*Alias)(c),
+	}
+	if err := json.Unmarshal(data, &a); err != nil {
```
byronharvey24, line 51: the address of a here would be the address of the address of the struct wouldn't it?

upalsaha (approved):
LG overall

entity_unmarshal.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"encoding/json"
+	"reflect"
+	"strings"
+)
+
+func UnmarshalEntityJSON(i interface{}, data []byte) error {
+	var jsonTagToKey = map[string]string{}
+	val := reflect.ValueOf(i).Elem()
+	for i := 0; i < val.Type().NumField(); i++ {
```
byronharvey24, line 12: can we not do a range function here?

byronharvey24 (approved):

customfield_service.go
byronharvey24: my surprise was that we didn't check both not that we did check the error!

customfield_service.go
cdworak: hah this is years old code (just moved to a new file) but thank you for your thoroughness

customfield_service.go
cdworak: i'm just going to leave as is because it's a copy paste!

customfield_service.go
cdworak: I think you could do either but the MultiOption type will be *[]string so I think it makes it easier to not have to do yext.Strings() and wrap the return. I'll explain a little more tomorrow in our training!

customfield_service.go
cdworak: same as above

customfield_service.go
cdworak: hmm this is a good point was a round about way of doing this

entity_diff.go
cdworak: yep yep

entity_diff.go
cdworak: no not assuming one occurance, the for loop will go as many times as necessary which will handle our double pointers

entity_diff.go
cdworak: no -- that's actually why when you write your tests in etl_go, that even though the location obj that gets created has more fields than what you define in your Want: yext.Location{City: "McLean"} your tests pass (without defining every other field also in your Want which would be a pain)

But that's how it's always been!

entity_diff_test.go
cdworak: to blank it out you need to send null  (we might want it be unspecified)

entity_diff_test.go
cdworak: same as above

entity_diff_test.go
cdworak: same as the reason you asked above: "do we not want to handle the reverse case if isZeroValue(valA) && !isZeroValue(valB)?"

entity_diff_test.go
cdworak: same as the answer to the question you asked above: "do we not want to handle the reverse case if isZeroValue(valA) && !isZeroValue(valB)?"

entity_service.go
cdworak: i think u is a url type not a string

entity_service.go
cdworak: yep at the top of the file, very standard for all of our services to have!

entity_service.go
cdworak: yep good call

entity_service.go
cdworak: lol could i have written a more round about way of returning this. no. 

good catch

entity_service.go
cdworak: lolol

entity_test.go
cdworak: this is actually standard for our string function

entity_test.go
cdworak: it's tricky because a wasn't definitely as a pointer

entity_service.go
cdworak: unfortunately yes i think. I think this has taken on a similar role to the CustomFieldManager of the CustomFieldService which is also exported. I had it not so at first but then in yext/etl ran into some trouble

entity_unmarshal.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"encoding/json"
+	"reflect"
+	"strings"
+)
+
+func UnmarshalEntityJSON(i interface{}, data []byte) error {
+	var jsonTagToKey = map[string]string{}
+	val := reflect.ValueOf(i).Elem()
```
trob-indie, line 11: indirect(reflect.ValueOf(i)) ?

entity_unmarshal.go
```diff
@@ -0,0 +1,40 @@
...
+
+	var m map[string]interface{}
+	err := json.Unmarshal(data, &m)
+	if err != nil {
+		return err
+	}
+
+	for tag, val := range m {
+		if _, ok := jsonTagToKey[tag]; ok && val == nil {
+			v := reflect.ValueOf(i).Elem().FieldByName(jsonTagToKey[tag])
```
trob-indie, line 26: Maybe I'm not clear on when you'd use indirect over .Elem(), but it seems safer to use here.

hours.go
```diff
@@ -11,6 +11,7 @@ const (
 	HoursClosedAllWeek = "1:closed,2:closed,3:closed,4:closed,5:closed,6:closed,7:closed"
 	HoursOpen24Hours   = "00:00:00:00"
 	HoursClosed        = "closed"
+	hoursLen           = 11 // XX:XX:XX:XX format
```
trob-indie, line 14: This comment seems unnecessary considering the value for HoursOpen24Hours

hours.go
```diff
@@ -178,55 +178,61 @@ func (h *HoursHelper) StringSerialize() string {
...
 	}
 	var d = &DayHours{}
+	intervals := []Interval{}
 	for _, interval := range h.GetHours(weekday) {
 		parts := strings.Split(interval, ":")
-		d.SetHours(fmt.Sprintf("%s:%s", parts[0], parts[1]), fmt.Sprintf("%s:%s", parts[2], parts[3]))
+		intervals = append(intervals,
+			Interval{
+				Start: fmt.Sprintf("%s:%s", parts[0], parts[1]),
+				End:   fmt.Sprintf("%s:%s", parts[2], parts[3]),
```
trob-indie, line 227: might want to check len(parts) > 3

hours.go
```diff
@@ -278,7 +284,14 @@ func parseWeekdayAndHoursFromString(str string) (Weekday, string, error) {
 	if err != nil {
 		return -1, "", fmt.Errorf("Error parsing weekday hours from string; unable to convert index to num: %s", err)
 	}
-	return Weekday(weekdayInt), strings.Join(hoursParts[1:], ":"), nil
+	if hoursParts[1] == HoursClosed {
```
trob-indie, line 287: len(hoursParts) > 1 && hoursParts[1] == HoursClosed ?

location.go
```diff
@@ -831,55 +831,17 @@ func (l LocationHolidayHours) String() string {
...
-
-type GoogleAttribute struct {
+type LocationGoogleAttribute struct {
 	Id        *string   `json:"id,omitempty"`
 	OptionIds *[]string `json:"optionIds,omitempty"`
 }
 
-func (g GoogleAttribute) String() string {
+func (g LocationGoogleAttribute) String() string {
 	return fmt.Sprintf("Id: '%v', OptionIds: '%v'", g.Id, g.OptionIds)
```
trob-indie, line 840: %s for g.Id?

entity_unmarshal.go
cdworak: the entity object will only ever be a pointer (like *yext.Location) so this should be safe!

PR #79, opened 2019-02-08
**Date to **string

cdworak:


entity_test.go
```diff
@@ -15,10 +15,11 @@ type CustomEntity struct {
 	CFPhoto        **Photo           `json:"cf_Photo,omitempty"`
 	CFVideos       *[]Video          `json:"cf_Videos,omitempty"`
 	CFVideo        **Video           `json:"cf_Video,omitempty"`
-	CFDate         **Date            `json:"cf_Date,omitempty"`
+	CFDate         *string           `json:"cf_Date,omitempty"`
 	CFSingleOption **string          `json:"cf_SingleOption,omitempty"`
 	CFMultiOption  *UnorderedStrings `json:"cf_MultiOption,omitempty"`
 	CFYesNo        **bool            `json:"cf_YesNo,omitempty"`
+	CFText         *string           `json:"cf_Text,omitempty"`
```
trob-indie, line 22: is this not a double pointer, too?

trob-indie (approved):

entity_test.go
cdworak: just a text field so "" will clear it!

PR #86, opened 2019-02-22
Add GetCountryCode() to LocationEntity

xzou:


byronharvey24 (approved):
Do we not have to check is BaseEntity is nil?

xzou:
@byronharvey24 I didn't because the other functions didn't, although if we run into an issue with it in the future then we should definitely update every function that uses BaseEntity!

PR #89, opened 2019-02-23
Return Hours reference from SetClosedAllWeek

bhainesva:
I would like to be able to do: 

```
LocationEntity: yext.LocationEntity{
  Hours: yext.NullableHours((&yext.Hours{}).SetClosedAllWeek()),
}
```

### DIFF
--- a/location_entity.go
+++ b/location_entity.go
@@ -309,6 +309,12 @@ func (h Hours) String() string {
 	return string(b)
 }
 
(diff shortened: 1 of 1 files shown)


mhupman (approved):
Lgtm, but would it make sense to have a static HoursClosedAllWeek() method that hides the pointer logic?  Then you would t need to worry about chaining.

PR #90, opened 2019-02-25
add healthcare_professional entity

cdworak:


healthcare_professional_entity.go
```diff
@@ -0,0 +1,515 @@
...
+// ** Phone: Send in 540-444-4444, get back 5404444444
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+const ENTITYTYPE_HEALTHCAREPROFESSIONAL EntityType = "healthcareProfessional"
+
+const ENTITYTYPE_HEALTHCAREPROFESSIONAL_CF EntityType = "HEALTHCARE_PROFESSIONAL"
```
mhupman: What does CF stand for?

healthcare_professional_entity.go
```diff
@@ -0,0 +1,515 @@
...
+
+// Location is the representation of a Location in Yext Location Manager.
```
mhupman: Comment is wrong

healthcare_professional_entity.go
```diff
@@ -0,0 +1,515 @@
...
+	YearCompleted   int    `json:"yearCompleted,omitempty"`
+}
+
+func (e Education) String() string {
+	return fmt.Sprintf("Institution Name: '%v', Type: '%v', Year Completed: '%v'", e.InstitutionName, e.Type, e.YearCompleted)
+}
+
+// Equal compares Education
+func (a *Education) Equal(b Comparable) bool {
+	defer func() {
```
mhupman: Is this panic likely to happen?

healthcare_professional_entity.go
mhupman: Basically if we don’t expect a panic normally it might be better to just let it bubble

mhupman (approved):

mhupman:
Oh, one last suggestion - I might consider naming the file entity_healthcare_professional.go.  That way alll the entity types sit together alphabetically.

healthcare_professional_entity.go
byronharvey24: CF stands for Custom Field, we have to support two separate API casings right now because of the custom fields create endpoint. There's an item in Apollo's backlog to fix this.

mhupman:
Yuck!

Sent from my iPhone

On Feb 25, 2019, at 9:58 AM, Byron <notifications@github.com> wrote:

*@byronharvey24* commented on this pull request.
------------------------------

In healthcare_professional_entity.go
<https://github.com/yext/yext-go/pull/90#discussion_r259861934>:

> @@ -0,0 +1,515 @@
+package yext
+
+// TODO
+// * Need better custom field accessors and helpers
+// * The API will accept some things and return them in a different
format - this makes diff'ing difficult:
+// ** Phone: Send in 540-444-4444, get back 5404444444
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+const ENTITYTYPE_HEALTHCAREPROFESSIONAL EntityType = "healthcareProfessional"
+
+const ENTITYTYPE_HEALTHCAREPROFESSIONAL_CF EntityType =
"HEALTHCARE_PROFESSIONAL"

CF stands for Custom Field, we have to support two separate API casings
right now because of the custom fields create endpoint. There's an item in
Apollo's backlog to fix this.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/yext/yext-go/pull/90#discussion_r259861934>, or mute
the thread

PR #92, opened 2019-02-25
Create review diff, update pointers, and add getter to entities

byronharvey24:


review.go
```diff
@@ -59,184 +61,266 @@ type ReviewLabel struct {
...
-func (y ReviewLabel) GetName() string {
+func (y *ReviewLabel) GetName() string {
 	if y.Name != nil {
 		return *y.Name
 	}
 	return ""
 }
+
+func (y *ReviewCreate) GetLocationId() string {
+	if y.LocationId != nil {
```
tbirdsall, line 247: Hmm, I wonder if this name will change in a new v param

review_diff.go
```diff
@@ -0,0 +1,37 @@
...
+	)
+
+	for i := 0; i < numA; i++ {
+		var (
+			nameA = aType.Field(i).Name
+			valA  = aValue.Field(i)
+			valB  = bValue.Field(i)
+		)
+
+		if valB.IsNil() || nameA == "Id" {
```
tbirdsall: Should this match "LocationId" ?

review_diff.go
```diff
@@ -0,0 +1,37 @@
+package yext
+
+import (
+	"reflect"
+)
+
+func (a *ReviewCreate) Diff(b *ReviewCreate) (d *ReviewCreate, diff bool) {
+	d = &ReviewCreate{LocationId: String(a.GetLocationId())}
+
+	var (
```
mhupman, line 10: is it possible to share this code from #10-31?  Doesn't seem like there is anything particular to ReviewCreate objects.

review.go
```diff
@@ -59,184 +61,266 @@ type ReviewLabel struct {
 	Name *string `json:"name"`
 }
 
-func (y Review) GetId() int {
+func (y *Review) GetId() int {
```
mhupman: Is it necessary to change all of these to pointer receivers?  Since they don't modify state could we leave them as is?

mhupman (approved):

review.go
byronharvey24: Sure!

review_diff.go
byronharvey24: I don't believe so because the LocationId is how the review gets associated with a given location but I could be wrong.

tbirdsall: but the ReviewCreate object doesn't have a value named "Id", right?

review_diff.go
```diff
@@ -20,7 +20,7 @@ func (a *ReviewCreate) Diff(b *ReviewCreate) (d *ReviewCreate, diff bool) {
 			valB  = bValue.Field(i)
 		)
 
-		if valB.IsNil() || nameA == "Id" {
+		if valB.IsNil() {
```
tbirdsall, line 23: Assuming we thoroughly checked this?

tbirdsall (approved):
LGTM!

review_diff.go
byronharvey24: Yup! I wrote lots of failing and succeeding tests to make sure I wasn't overlooking something.

PR #93, opened 2019-02-25
entity_diff: prevent diff from returning too early

cdworak:


upalsaha (approved):
LGTM, but could you also add these unit tests? I think this might solve the problem I was seeing for CFT assets earlier https://github.com/yext/yext-go/commit/e9842c4aee87e2ad509fb2aff32e5853a09adf7b

PR #95, opened 2019-02-27
Add methods to DayHours; null check *Hours in GetDayHours

xzou:


entity_location.go
```diff
@@ -341,6 +341,20 @@ func (d *DayHours) SetClosed() {
 	d.OpenIntervals = nil
 }
 
+func (d *DayHours) Closed() bool {
+	if d == nil {
```
cdworak: hi @xzou is this assumption correct?

My concern here is that if you send &DayHours{} to the API, it doesn't actually close the hours for that day. You have to send &DayHours{IsClosed: true}

mhupman (commented):
How could the receiver be nil in practice?

cdworak:
I think it's easy to stumble across one accidentally. If you have h := &Hours{} and you call h.Monday.IsClosed() that would be an example

mhupman:
Ah, interesting.  Wonder if we could pursue an alternate API, e.g. `hours.IsClosed(yext.Monday)` so you wouldn't have to worry about missing DayHours.  We could also make them not pointers so they can't be missing.  Thanks for the background!

entity_location.go
xzou: @cdworak would it make sense then for this method to just return d.IsClosed, and for the null-checking to take place on the outside? since the logic pertaining to what to do with a nil DayHours probably varies by caller

cdworak:
Agreed, I think that func (h *Hours) IsClosed(w *Weekday) would be a safer alternative. @xzou what do you think?

xzou:
@cdworak I can absolutely change it to that! Although we'd still have to check whether or not the hours for the day passed into `hours.IsClosed(day)` is nil since we'd still have to get those hours from within the method before accessing its `IsClosed` attribute right?

cdworak (approved):

PR #98, opened 2019-03-01
Update type.go

bhainesva:
Can we do this


cdworak:
Per discussion offline, I think if we need to handle a nil then this should return nil rather than []string{}

PR #99, opened 2019-03-01
add partial hours support

byronharvey24:
I am really hesitant to mess with the locationhourshelper but this could be refactored to reduce code repetition. I just want to provide support for non-closing hours.

### DIFF
--- a/hours.go
+++ b/hours.go
@@ -268,8 +268,13 @@ func (h *LocationHoursHelper) StructSerializeDay(weekday Weekday) *DayHours {
 }
 
 func (h *LocationHoursHelper) StructSerializeDayPartial(weekday Weekday) *DayHours {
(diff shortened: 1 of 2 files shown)


mhupman:
What do you mean by non-closing hours?  Does this mean we need two different ways of parsing hours?

byronharvey24:
Yeah we do need two different ways is what I am saying. So for example when a client gives hours for Tuesday and Wednesday only there are two potential outcomes: 1. We consider all other days closed (current implementation of struct serialization and what many clients opt for) or 2. We leave the other days untouched and only update hours for the Tuesday and Wednesday. My code solves for the second scenario while leaving the first unchanged.

bhainesva:
This looks reasonable to me, I feel like this makes  more sense as the default behavior (not assuming days without information are closed) than what we're doing currently but I'll let @cdworak decide if that's something we actually want to update. 

+1

byronharvey24:
After conversation with Catherine this is not supported by the locations API so we will not pursue this change.

PR #100, opened 2019-03-04
add entity ACL function to users_service

alexisgrow:


user_service.go
```diff
@@ -119,3 +119,12 @@ func (u *UserService) NewLocationACL(l *Location, r Role) ACL {
 		AccessOn:  ACCESS_LOCATION,
 	}
 }
+
+func (u *UserService) NewEntityACL(e Entity, r Role) ACL {
```
byronharvey24, line 123: I would prefer e to be a pointer to an Entity!

byronharvey24 (approved):

user_service.go
```diff
@@ -119,3 +119,12 @@ func (u *UserService) NewLocationACL(l *Location, r Role) ACL {
...
+    return ACL{
+        Role:      r,
+        On:        e.GetEntityId(),
+        AccountId: u.client.Config.AccountId,
+        AccessOn:  ACCESS_LOCATION,
```
tbirdsall, line 128: Do we need to change this? 

`ACCESS_LOCATION = AccessOn("LOCATION")` -> is that okay to be location?

tbirdsall (approved):
LGTM

user_service.go
alexisgrow: As discussed offline, this isn't a great option since Entity is an interface! Here, we'd pass in a pointer to the implementation of the interface (e.g. farmers.FarmersLocation) rather than a pointer to the interface itself.

user_service.go
alexisgrow: Yeah asked Commando and they have not moved it yet, I'm not sure if it was on their radar before, but I'm sure (err, I hope) they'll tell us when it changes!

PR #102, opened 2019-03-06
Update hours helper struct serialize to return NullRegularHours()

xzou:


type.go
```diff
@@ -197,6 +197,20 @@ func NullHours() **Hours {
 	return &v
 }
 
+func NullRegularHours() **Hours {
+	var v *Hours
+	v = &Hours{
+		Sunday:    NullableDayHours(nil),
```
cdworak: there's a NullDayHours() in entity_location.go, can we use that here?

xzou: oh that's perfect! updated

cdworak (approved):

PR #105, opened 2019-03-13
Add delete operation to entity service

byronharvey24:


entity_service.go
```diff
@@ -267,3 +267,13 @@ func (e *EntityService) EditWithId(y Entity, id string) (*Response, error) {
 func (e *EntityService) Edit(y Entity) (*Response, error) {
 	return e.EditWithId(y, y.GetEntityId())
 }
+
+// DeleteWithId sends a request to the Knowledge Entities API to delete an entity with a given id
+func (e *EntityService) DeleteWithId(y Entity, id string) (*Response, error) {
```
cdworak: this should just take an Id (no y Entity) and you should just need the one function Delete(id string)

tbirdsall (approved):
LETM

PR #107, opened 2019-03-18
add support for LanguageProfile:ListAll

bhainesva:


language_profile_service.go
```diff
@@ -60,6 +72,77 @@ func (l *LanguageProfileService) List(id string) ([]*Entity, *Response, error) {
...
+		return resp.PageToken, nil
+	}
+
+	if err := tokenListHelper(lg, &opts.ListOptions); err != nil {
+		return nil, err
+	}
+	return entities, nil
+}
+
+func (l *LanguageProfileService) ListAll(opts *EntityListOptions) (*LanguageProfileListAllResponse, *Response, error) {
```
cdworak: can you make this e or l to match the one above?

cdworak: would anyone ever call ListAll and not ListAllPaginated? If not, should we not export it?

cdworak: I'd love to swap the names on this to match what we've historically done. ListAll being the paginated one. Maybe ListHelper or ListPage or ListSingle for the single page one?

bhainesva: For entities, the name of the endpoint is List and we created a ListAll function on top of it. 
Here, the actual endpoint is ListAll, it feels misleading to me to have ListAll do extra things on top of the actual endpoint. But sure, can change.

cdworak (approved):

PR #112, opened 2019-03-26
entity_unmarshal: handle null day hours (and other null values in obj…

cdworak:
…ects)


entity_unmarshal.go
```diff
@@ -6,35 +6,45 @@ import (
...
-	if err != nil {
-		return err
-	}
-
 	for tag, val := range m {
-		if _, ok := jsonTagToKey[tag]; ok && val == nil {
-			v := reflect.ValueOf(i).Elem().FieldByName(jsonTagToKey[tag])
+		if _, ok := jsonTagToKey[tag]; ok {
+			if val == nil {
+				v := Indirect(reflect.ValueOf(i)).FieldByName(jsonTagToKey[tag])
```
byronharvey24, line 21: Why do we need to Indirect here?

byronharvey24 (approved):
This looks good to me. I'm not great with reflect so it would be helpful to me if there were some comments here

entity_unmarshal.go
cdworak: you can't call FieldByName on a pointer, so Indirect will remove remove a pointer if it exists!

PR #113, opened 2019-03-28
entity_unmarshal: Check if type is struct before calling NumField()

xzou:


entity_unmarshal.go
```diff
@@ -9,10 +9,12 @@ import (
 func unmarshal(i interface{}, m map[string]interface{}) interface{} {
 	var jsonTagToKey = map[string]string{}
 	val := Indirect(reflect.ValueOf(i))
-	for i := 0; i < val.Type().NumField(); i++ {
-		field := val.Type().Field(i)
-		tag := strings.Replace(field.Tag.Get("json"), ",omitempty", "", -1)
-		jsonTagToKey[tag] = field.Name
+	if val.Kind() == reflect.Struct {
```
cdworak, line 11: this if statement should probably wrap everything i think in this function, because the part you wrapped generates jsonTagToKey and that part is used in the second part (so if we don't actually populate jsonTagToKey there's no use in doing this part.

Do you have an example for which this was breaking? Should we add something to the entity_test.go file?

xzou: yes that makes sense--I've updated it! I've also added google attributes to the test. prior to my fix, adding google attributes resulted in a panic when trying to run the tests

cdworak (approved):

PR #117, opened 2019-04-10
Degrees Fix

pshah28:


byronharvey24 (approved):
Is there a getter or setter associated with this that needs to be changed?

trob-indie (approved):

PR #121, opened 2019-04-17
entity_diff: handle double pointer to nil (aka null)

cdworak:


trob-indie (approved):
Love it! Wish I had it for Ohio health :/

PR #122, opened 2019-04-25
unordered strings equal: Update null checking for interface

alexisgrow:


upalsaha (approved):
LGTM, but @cdworak thoughts? Feel like if the assertion fails here it would be desirable for it to panic so this should be fine

upalsaha:
Ah yeah, we're recovering anyway with context so @alexisgrow you're good to merge

PR #125, opened 2019-04-29
type: add date pointer

tbirdsall:


type.go
```diff
@@ -263,3 +263,9 @@ func NullableUnorderedStrings(v []string) *UnorderedStrings {
 	u := UnorderedStrings(v)
 	return &u
 }
+
```
cdworak, line 266: We stopped using the location custom field types that were just wrappers (SingleLineText is just *String) with the entity migration, as the custom types are no longer needed since CustomFields is no longer a map[string]interface. We've just been using **String (because if you want to blank it out you need to send null, hence the double point). If you would like to use Date I'd recommend returning **Date so you can null it out

tbirdsall: 👍 will just change to **String and abandon this PR. Don't think there's any advantage to using the Date wrapper in my case

PR #128, opened 2019-05-10
SetLabelsWithUnorderedStrings: handle nil J=PC-43989

cdworak:


entity_test.go
```diff
@@ -154,6 +154,91 @@ func TestEntityJSONDeserialization(t *testing.T) {
...
+				BaseEntity: BaseEntity{
+					Meta: &EntityMeta{
+						Labels: nil,
+					},
+				},
+			},
+			Source: &LocationEntity{
+				BaseEntity: BaseEntity{
+					Meta: &EntityMeta{
+						Labels: nil,
```
pshah28, line 174: might be worth adding a test with labels populated just for posterity

pshah28 (approved):

PR #129, opened 2019-05-20
Image comparison: use source url

cdworak:


entity_location.go
```diff
@@ -226,6 +228,26 @@ func NullImage() **Image {
 	return &i
 }
 
+// If source URL is filled in, Image came from Yext API and source URL should be
+// used for comparison as URL is the mktg cdn generated URL
+func (a *Image) Equal(b Comparable) bool {
+	defer func() {
+		if r := recover(); r != nil {
+			log.Printf("Value of A: %+v, Value of B:%+v, Type Of A: %T, Type Of B: %T\n", a, b, a, b)
```
mhupman, line 236: mean to leave this in here?

entity_location.go
```diff
@@ -226,6 +228,26 @@ func NullImage() **Image {
...
+			panic(r)
+		}
+	}()
+
+	if a.SourceUrl != nil {
+		a.Url = a.SourceUrl
```
mhupman, line 242: mutating the source seems wacky - can we compare the source URL directly?

mhupman (commented):
Actually I'm not so sure this is desirable - this only works if the source image urls change every time there is a change to the image.  In an ETL like ------- Headshot, this wouldn't work because their URLs don't change even if the underlying images do.

entity_location.go
cdworak: sure but i wanted to make sure it was only the source url and url that is different, i don't have a great way of doing that other than porting reflection here (the diff) or writing code to compare attribute to attribute, which if we add an attribute we'd have to remember to update this code too, so that seems dangerous

cdworak:
I think if ------- doesn't update their source than we should probably add a vparam to the end of the URL with today's date and we'd essentially have no change of functionality.

For clients that do update their source we get duplicate updates, which messes with stats and fills the db unnecessarily, so I'd prefer the vparam route if possible.

entity_location.go
cdworak: Yeah, it follows the model of the other Equal implementations in yext-go

entity_location.go
bhainesva, line 242: I'm uncomfortable with the idea of the equal function modifying the receiver that it's called on? could we copy it instead and then call genericdiff on the copy?

cdworak: sure!

mhupman:
> I think if -------- doesn't update their source than we should probably add a vparam to the end of the URL with today's date and we'd essentially have no change of functionality.
> 
> For clients that do update their source we get duplicate updates, which messes with stats and fills the db unnecessarily, so I'd prefer the vparam route if possible.

You mean modify their source URL?  How so?

bhainesva:
Just want to add another use case, ---- also has an etl that updates images without updating the url

mhupman:
We could make this opt-in (or out) by using a property on the image itself?  e.g. `image.SetUnpredictableImageContents(true)`

benmcginnis:
Another potential avenue would be to add a hash property on the image itself and use that for diffing if it's populated.  That would allow us to have calling code handle it by downloading the image and creating a hash of the contents to diff against for cases where we don't get changed URLs and when we do get changed URLs we can just set the hash to the url itself.

PR #133, opened 2019-07-09
Add and register ATM entity

xzou:


entity_atm.go
```diff
@@ -0,0 +1,334 @@
+package yext
+
+import (
+	"encoding/json"
+)
+
+const (
+	ENTITYTYPE_ATM    EntityType = "atm"
+	ENTITYTYPE_ATM_CF EntityType = "ATM"
```
cdworak: can you check to see if we still need this? @byronharvey24 should have more context

cdworak (approved):

entity_atm.go
xzou: @cdworak looks like they updated it, but we're going to have to bump up our v param! I shall look into it

PR #136, opened 2019-07-17
add healthcare facility entity

pshah28:


entity_healthcare_facility.go
```diff
@@ -0,0 +1,523 @@
+package yext
+
+// TODO
```
cdworak: can we remove this todo? i think we have it in another file so it's fine

entity_healthcare_facility.go
```diff
@@ -0,0 +1,523 @@
...
+import (
+	"encoding/json"
+)
+
+const ENTITYTYPE_HEALTHCAREFACILITY EntityType = "healthcareFacility"
+
+// Entity types for custom fields must be MACRO_CASED
+// TODO: wait for completion of this item: https://yexttest.atlassian.net/browse/AO-3660 (ETC End of February 2019)
+// After item is copmleted up the vparam and (hopefully) delete this line and remove all references to it.
+const ENTITYTYPE_HEALTHCAREFACILITY_CF EntityType = "HEALTHCARE_FACILITY"
```
cdworak: i believe you can also delete this and the TODO (confirm with Cindy but I'm pretty sure she just pushed an item to remove)

cdworak (approved):

PR #138, opened 2019-07-22
Add and register Hotel entity

xzou:


entity_hotel.go
```diff
@@ -0,0 +1,509 @@
...
+	TtyPhone       *string   `json:"ttyPhone,omitempty"`
+	Emails         *[]string `json:"emails,omitempty"`
+
+	// Location Info
+	AdditionalHoursText *string   `json:"additionalHoursText,omitempty"`
+	Associations        *[]string `json:"associations,omitempty"`
+	Brands              *[]string `json:"brands,omitempty"`
+	Description         *string   `json:"description,omitempty"`
+	Hours               **Hours   `json:"hours,omitempty"`
+	CheckInTime         *string   `json:"checkInTime,omitempty"` // TODO: check type of time field
```
trob-indie, line 40: Still TODO?

trob-indie (approved):

entity_hotel.go
xzou: ah yup! I won't merge this until after QA probably. this field was in the original JSON spec they sent but wasn't in the account, so I need to follow up about its status (same w/ CheckOutTime)

PR #139, opened 2019-07-31
Add getter for Date and nullable Date

the-nemz:


the-nemz:
Closing due to offline conversation!

PR #140, opened 2019-08-01
Added custom field type tests to entity differ

the-nemz:


the-nemz:
Two of these tests currently fail
- line 2237: trying to clear the Image subfield of the CFT
- line 2267: trying to clear the whole CFT

trob-indie (approved):
lgtm

PR #142, opened 2019-08-27
Raw entity

cdworak:
The current yext.Diff() function cannot handle a map, only a struct. After attempting to add map functionality the yext.Diff() in the interest of time and simplicity I decided to write something specific for the RawEntity (which is a map). In the future, would be nice to have yext.Diff() automatically handle a map but would require a significant lift as the for loop within yext.Diff is entirely specific to a struct

### DIFF
--- a/entity_diff.go
+++ b/entity_diff.go
@@ -202,9 +202,9 @@ func RawEntityDiff(a map[string]interface{}, b map[string]interface{}, nilIsEmpt
 			_, aIsMap := aVal.(map[string]interface{})
 			_, bIsMap := aVal.(map[string]interface{})
 			if aIsMap && bIsMap {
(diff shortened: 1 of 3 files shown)


entity_diff.go
```diff
@@ -167,9 +167,56 @@ func Diff(a Entity, b Entity) (Entity, bool, error) {
...
+		isDiff = false
+	)
+
+	for key, bVal := range bAsMap {
+		if key == "nilIsEmpty" {
+			continue
+		}
+		if aVal, ok := aAsMap[key]; ok {
+			_, aIsMap := aVal.(map[string]interface{})
+			_, bIsMap := aVal.(map[string]interface{})
```
pshah28: bVal.(map[string]interface{})

pshah28 (changes requested):

entity_diff.go
cdworak: yikes

cdworak: thanks!

trob-indie (approved):
lgtm

PR #160, opened 2019-10-31
added more fields to reviews struct

AidanFitzgerald:


review.go
```diff
@@ -23,13 +24,16 @@ type Review struct {
 	AuthorEmail        *string        `json:"authorEmail"`
 	URL                *string        `json:"url"`
 	PublisherDate      *int           `json:"publisherDate"`
-	LastYextUpdateDate *int           `json:"lastYextUpdateDate"`
```
AidanFitzgerald, line 26: The name of this must've changed in the API, API uses "Time" not "Date". See here: https://developer.yext.com/docs/api-reference/#operation/getReview

I double checked that no ETL is currently using this, so this update won't break anything

bhainesva: I'd check w/ @cdworak on how she wants to handle, but since this repo is open source it's probably not enough to check our own code for usages.

tbirdsall: @AidanFitzgerald so it's a version thing? As in, if you pass an older v param with a get request, can you get back lastYextUpdateDate? Or was it always incorrect in our code

AidanFitzgerald: @tbirdsall It looks like it was always "Time" in the API, maybe it was just always incorrect in our code?

AidanFitzgerald: Tried a v param with a year set to 2010 and it was still lastYextUpdateTime, so probably safe to update because anyone that's using lastYextUpdateDate it won't have done anything anyways, right?

AidanFitzgerald: ^ Ben pointed out that this isn't a safe assumption because even if their code never did anything updating this will make their references incorrect and cause problems

cdworak (approved):

PR #161, opened 2019-11-01
Fixing bicycle values

Zavoky:


pshah28 (approved):
lgtm. just wondering, are there any ETLs that currently use the field?

Zavoky:
Nope, IHG / other hotel etls don't use this field, just Choice which i'm working on

PR #172, opened 2019-12-17
raw entity: get and set helpers

cdworak:


client.go
```diff
@@ -208,10 +208,6 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 		// Rehydrate the request body since it might have been drained by the previous attempt
 		req.Body = ioutil.NopCloser(bytes.NewBuffer(originalRequestBody))
 
-		if c.Config.Logger != nil {
```
cdworak, line 211: i deleted this because the logging was rather noisy (it prints the whole body of the request)

pshah28 (approved):
lgtm

PR #177, opened 2020-01-11
Adding EntityList

Zavoky:


customfield.go
```diff
@@ -35,3 +37,27 @@ func (c CustomField) GetId() string {
 	}
 	return *c.Id
 }
+
+type EntityList UnorderedStrings
```
cdworak, line 41: hmm do we need a specific type for this? Could we just use UnorderedStrings (like we do for multi-option)

the reason I ask is because we've moved away from custom types (each custom field used to have a custom type, like single option, multi option) and we've simplified that to be the underlying type (like string and unordered strings). Also I could imagine that in some cases the order matters to an entity list (like for Pages, the display order might matter). I understand that to the linker it doesn't matter so much but there could be a case in which it does!

Zavoky: fair, I'll use unorderedstrings for the linker instead

PR #187, opened 2020-03-09
-J=PC-63853 changes to entity_registry and unmarshal needed for hybri…

AidanFitzgerald:
…d history mode

### DIFF
--- a/entity_unmarshal.go
+++ b/entity_unmarshal.go
@@ -23,22 +23,20 @@ func unmarshal(i interface{}, m map[string]interface{}) interface{} {
 					v := Indirect(reflect.ValueOf(i)).FieldByName(jsonTagToKey[tag])
 
 					// Check if double pointer
-					if v.Type().Kind() == reflect.Ptr {
-						t := v.Type()
-						if t.Kind() == reflect.Ptr && t.Elem().Kind() == reflect.Ptr {
-							for t.Kind() == reflect.Ptr && t.Elem().Kind() == reflect.Ptr {
-								t = t.Elem()
-							}
-							typedNil := reflect.New(t)
-
-							defer func() {
-								if r := recover(); r != nil {
-									log.Fatalf("Error while unmarshaling field '%s': %s", jsonTagToKey[tag], r)
-								}
-							}()
-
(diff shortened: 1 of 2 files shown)


entity_unmarshal.go
```diff
@@ -25,18 +25,20 @@ func unmarshal(i interface{}, m map[string]interface{}) interface{} {
...
 						t := v.Type()
-						for t.Kind() == reflect.Ptr && t.Elem().Kind() == reflect.Ptr {
-							t = t.Elem()
-						}
-						typedNil := reflect.New(t)
-
-						defer func() {
-							if r := recover(); r != nil {
-								log.Fatalf("Error while unmarshaling field '%s': %s", jsonTagToKey[tag], r)
+						if t.Kind() == reflect.Ptr && t.Elem().Kind() == reflect.Ptr {
```
cdworak: ah i finally figured out why this looked so strange. Instead of 

if v.Type().Kind() == reflect.Ptr {
   t := v.Type()							
    if t.Kind() == reflect.Ptr && t.Elem().Kind() == reflect.Ptr {
...
We can just do

if v.Type().Kind() == reflect.Ptr && v.Type().Elem().Kind() == reflect.Ptr {

}

and only have one if (t is just equal to v.Type()	which was what was confusing me!)

AidanFitzgerald: do we want to get rid of the declaration of t? it's used a decent amount (and gets reassigned multiple times in the for loop to t.Elem()), not sure if it would cause problems to reassign v.Type() instead of having a separate variable

maybe we can just remove the redundant first part of the second if statement since we've already checked if v.Type().Kind() == reflect.Ptr?:

```
if v.Type().Kind == reflect.Ptr {
    t := v.Type()
    if t.Elem().Kind() == reflect.Ptr {
    ...
}

cdworak: Sure you can keep the t declaration but we should consolidate the if statements

so instead of 
if v.Type().Kind() == reflect.Ptr {
						t := v.Type()
						if t.Kind() == reflect.Ptr && t.Elem().Kind() == reflect.Ptr {
... everything else

it would be:
t := v.Type()
if t.Kind() == reflect.Ptr && t.Elem().Kind() == reflect.Ptr {
... everything else

cdworak (approved):

PR #195, opened 2020-05-06
Analytics fields

Zavoky:


analytics_service.go
```diff
@@ -10,6 +10,7 @@ type AnalyticsFilters struct {
 	StartDate                  *string   `json:"startDate"`
 	EndDate                    *string   `json:"endDate"`
 	LocationIds                *[]string `json:"locationIds"`
+	EntityTypes                *[]string `json:"entityType"`
```
jesuyedavid, line 13: should this be `json:"entityTypes"` instead of `json:"entityType"`

Zavoky: yep that was intentional

jesuyedavid (approved):
lgtm

PR #200, opened 2020-08-05
Hotel and healthcare updates

nirmalpatel94:


entity_healthcare_professional.go
```diff
@@ -338,6 +339,14 @@ func (y HealthcareProfessionalEntity) GetReservationUrl() string {
 	return ""
 }
 
+func (y HealthcareProfessionalEntity) GetTelehealthUrl() string {
```
pshah28, line 350: i think this should return a yext.Website{} in which you should be able to access the Url field without a nil pointer

pshah28 (changes requested):

pshah28 (approved):

PR #209, opened 2020-12-29
fix only deltas being sent for a field sync

nirmalpatel94:


entity_diff.go
```diff
@@ -204,7 +204,7 @@ func RawEntityDiff(a map[string]interface{}, b map[string]interface{}, nilIsEmpt
 			if aIsMap && bIsMap {
 				subFieldsDelta, subFieldsAreDiff := RawEntityDiff(aVal.(map[string]interface{}), bVal.(map[string]interface{}), nilIsEmptyA, nilIsEmptyB)
```
jsa3qy, line 205: get rid of subFieldsDelta

PR #214, opened 2021-04-06
feat(entity_service): add rich text format field

nirmalpatel94:
TEST=manual

### DIFF
--- a/entity_service.go
+++ b/entity_service.go
@@ -12,6 +12,28 @@ const (
 	EntityListMaxLimit = 50
 )
 
+type RichTextFormat int
+
+const (
+	RichTextFormatDefault RichTextFormat = iota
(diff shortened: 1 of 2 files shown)


entity_service.go
```diff
@@ -12,6 +12,27 @@ const (
...
+func (r RichTextFormat) ToString() string {
+	switch r {
+	case RichTextFormatHTML:
+		return "html"
+	case RichTextFormatMarkdown:
+		return "markdown"
+	case RichTextFormatNone:
+		return "none"
+	}
+	return ""
```
danielwuhoo: should we make this a default case instead just for cleanliness

danielwuhoo (approved):

pshah28 (approved):

PR #220, opened 2021-05-25
adds helpArticle entity type

nsiskind:


entity_helparticle.go
```diff
@@ -0,0 +1,52 @@
...
+	BaseEntity
+
+	// Admin
+	Keywords *[]string `json:"keywords,omitempty"` // LIST
+
+	// Content
+	Name             *string `json:"name,omitempty"`             // STRING
+	Body             *string `json:"body,omitempty"`             // RICH_TEXT
+	ShortDescription *string `json:"shortDescription,omitempty"` // STRING
+	VoteCount        *int    `json:"voteCount,omitempty"`        // STRING
```
AidanFitzgerald, line 19: are these actually ints in the JSON? feel like with number fields we usually just use a string but I do see some ints in the hotel entity type. they also use a double pointer in that entity def but i don't think that's necessary?

entity_helparticle.go
```diff
@@ -0,0 +1,52 @@
...
+	VoteSum          *int    `json:"voteSum,omitempty"`          // STRING
+	Promoted         **bool  `json:"promoted,omitempty"`         // BOOl
+
+	ExternalArticlePostDate   *Date `json:"externalArticlePostDate,omitempty"`
```
AidanFitzgerald, line 23: where is Date defined? can't find it lol. For date fields I think we also usually use a *string but could be wrong here

entity_helparticle.go
```diff
@@ -0,0 +1,52 @@
...
+	// Knowledge Assistant
+	PrimaryConversationContact *string `json:"primaryConversationContact,omitempty"`
+	NudgeEnabled               **bool  `json:"nudgeEnabled,omitempty"`
+
+	// Internal Use
+	Folder   *string `json:"folder,omitempty"`
+	Timezone *string `json:"timezone,omitempty"`
+}
+
+func (j *HelpArticle) String() string {
```
AidanFitzgerald: prob switch this to h

entity_helparticle.go
nsiskind: date is in location_customfield.go ... its just a string though. like this

type Date string

entity_helparticle.go
nsiskind: https://api-sandbox.yext.com/v2/accounts/me/entities/123?api_key=44d8cc0586f38db2503c8203ccf4e0ec&v=20200625

it does come back as an int ... i am not sure whether i should use a double pointer. not sure its necessary... that is when we want to send nil objects right?

entity_helparticle.go
nsiskind: 👍

AidanFitzgerald (approved):

PR #223, opened 2021-06-17
Add Activity Logs and update some Analytics Report

zxu10:
J=PC-134868

### DIFF
--- a/activity_log_service.go
+++ b/activity_log_service.go
@@ -1,0 +1,49 @@
+package yext
+
+const (
+	activityLogPath = "analytics/activity"
+
+	// Use the following constants for Activity Filter
+	ACTIVITYTYPES_LOCATION_UPDATED              = "LOCATION_UPDATED"
+	ACTIVITYTYPES_PUBLISHER_SUGGESTION_CREATED  = "PUBLISHER_SUGGESTION_CREATED"
+	ACTIVITYTYPES_PUBLISHER_SUGGESTION_APPROVED = "PUBLISHER_SUGGESTION_APPROVED"
+	ACTIVITYTYPES_PUBLISHER_SUGGESTION_REJECTED = "PUBLISHER_SUGGESTION_REJECTED"
+	ACTIVITYTYPES_REVIEW_CREATED                = "REVIEW_CREATED"
+	ACTIVITYTYPES_SOCIAL_POST_CREATED           = "SOCIAL_POST_CREATED"
+	ACTIVITYTYPES_LISTING_LIVE                  = "LISTING_LIVE"
+	ACTIVITYTYPES_DUPLICATE_SUPPRESSED          = "DUPLICATE_SUPPRESSED"
+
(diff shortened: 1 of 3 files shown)


activity_log_service.go
```diff
@@ -0,0 +1,17 @@
...
+	client *Client
+}
+
+type ActivityLogResponse struct {
+	Data []*ActivityLog `json:"activities"`
+}
+
+func (a *ActivityLogService) Create() (*ActivityLogResponse, *Response, error) {
+	arr := &ActivityLogResponse{}
+	r, err := a.client.DoRequestJSON("POST", activityLogPath, nil, arr)
```
AidanFitzgerald: i think even if you don't use them for this ETL it would be helpful to add support for the filters in the API docs so that others can use them in the future! I think you'd just want to make a struct for like "ActivityLogRequest" with fields for limit, offset, filters and then also a struct for a the filters (startDate, endDate, locationIds, activityTypes, actors), similar to what we're doing in the analytics_service.go file!

zxu10: updated!

AidanFitzgerald (commented):
looks great! sorry one last thing I forgot in my last post, I think it could be helpful as well to add consts for the values that are allowed in the activityTypes and actors fields so people don't have to double check the API docs to know what's allowed there! so like

```
const (
   ACTIVITYTYPES_LOCATION_UPDATED = "LOCATION_UPDATED"
   ACTIVITYTYPES_PUBLISHER_SUGGESTION_CREATED = "PUBLISHER_SUGGESTION_CREATED"
   ...
   ACTORS_YEXT_SYSTEM = "YEXT_SYSTEM"
   ... etc!
)
```

zxu10:
> looks great! sorry one last thing I forgot in my last post, I think it could be helpful as well to add consts for the values that are allowed in the activityTypes and actors fields so people don't have to double check the API docs to know what's allowed there! so like
> 
> ```
> const (
>    ACTIVITYTYPES_LOCATION_UPDATED = "LOCATION_UPDATED"
>    ACTIVITYTYPES_PUBLISHER_SUGGESTION_CREATED = "PUBLISHER_SUGGESTION_CREATED"
>    ...
>    ACTORS_YEXT_SYSTEM = "YEXT_SYSTEM"
>    ... etc!
> )
> ```

zxu10:
@AidanFitzgerald Looks like Yext API does not list all of the possible fields for activity types and actors...? Adding only a portion of what's available might be confusing.

AidanFitzgerald:
> @AidanFitzgerald Looks like Yext API does not list all of the possible fields for activity types and actors...? Adding only a portion of what's available might be confusing.

oh that's annoying! so you're saying there are options besides these? 
image

I think I'd still prefer us add the ones we do know about, maybe with a comment saying "there are other options not in the documentation, if you need one not listed here please add it!" or something like that?

zxu10:
@AidanFitzgerald You are right - I missed the dropdown menu on the API docs! Adding these constants now.

AidanFitzgerald (approved):

PR #229, opened 2021-07-07
healthcare professional: implements setValidDegrees method

jsa3qy:
- Checks degrees against hardcoded list of valid degrees supported by our platform
- sets degrees on self, returns parse error if any are invalid

J=PC-113595
TEST=manual

### DIFF
--- a/entity_healthcare_professional.go
+++ b/entity_healthcare_professional.go
@@ -8,6 +8,7 @@ package yext
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


entity_healthcare_professional.go
```diff
@@ -559,3 +560,34 @@ func (y HealthcareProfessionalEntity) GetCategoryIds() (v []string) {
 	}
 	return v
 }
+
+func (y *HealthcareProfessionalEntity) SetValidDegrees(degrees []string) []string {
```
AidanFitzgerald, line 567: love it! two small style-guidey things:

1. can you move invalidDegrees above validDegrees in the var block so it's alpha
2. can you add a short function comment just to make it clear that what gets returned are the _invalid_ degrees? i think returning them and then handling the parseError in the ETL is a prefect solution, but slightly unintuitive to return the bad values so having a comment alleviates that imo

AidanFitzgerald (approved):

PR #252, opened 2021-11-03
Add analytics clicks event and update page views event

stephh2:


analytics_data.go
```diff
@@ -27,7 +27,7 @@ type AnalyticsData struct {
 	AverageRating                         *float64 `json:"Average Rating"`
 	NewReviews                            *int     `json:"Reviews"`
 	StorepagesSessions                    *int     `json:"Storepages Sessions"`
-	StorepagesPageviews                   *int     `json:"Page Views"`
+	StorepagesPageviews                   *int     `json:"Pages Views"`
```
bhainesva, line 30: sorry I don't have much context, did this change or was the old value just wrong? 

if they changed the name behind a v param or something we would need to handle being able to unmarshal either version right?

stephh2: It didn't change This was just wrong from get go I believe

bhainesva (approved):

PR #254, opened 2021-11-04
feat: add service for admin apis

danielwuhoo:


Zavoky (commented):
we'll want to do a rename to account_service, since the administrative api encompasses a few more services

Zavoky (approved):

PR #267, opened 2022-02-08
add Entity struct to Reviewer struct

cblair28:


review.go
```diff
@@ -3,13 +3,18 @@ package yext
...
+	FirstName  *string       `json:"firstName,omitempty"`
+	LastName   *string       `json:"lastName,omitempty"`
+	Contact    *string       `json:"contact,omitempty"`
+	Image      *bool         `json:"image,omitempty"`
+	TemplateId *string       `json:"templateId,omitempty"`
+	LabelIds   []*string     `json:"labelIds,omitempty"`
+}
+
+type ReviewEntity struct {
+	Id *string `json:"id,omitempty"`
```
bhainesva: does this need to be a pointer / can it ever be empty?

bhainesva (approved):

review.go
cblair28: this needs to be a pointer because of how the review_service.go is setup, and it can be empty if the v param is less than 20210728

bhainesva: oh interesting I would have assumed there was no `entity` in that case rather than an entity with an empty id

cblair28: Actually, now that I think about it, you might be right on both accounts here - going to update and do some testing locally

cblair28: Yup - after testing, confirming that this should be a string instead of string pointer, and we don't need the 'omitempty' keyword. Thanks!

PR #270, opened 2022-03-16
add rendered option for entity language profiles

mtusman:
For alternative language profiles, we need to pass `rendered=true` to retrieve resolved embedded fields. Currently, we're using `resolvePlaceholder` in [entity.go](https://gerrit.yext.com/plugins/gitiles/congo/+/refs/heads/master/yext/account/entity.go#67), which is only valid parameter for the `/entities` endpoint, not `/entityprofiles`.

### DIFF
--- a/entity_service_test.go
+++ b/entity_service_test.go
@@ -57,6 +57,7 @@ func TestEntityListOptions(t *testing.T) {
 		searchIDs           string
 		entityTypes         string
 		resolvePlaceholders bool
+		rendered            bool
 		filter              string
 		format              string
 	}{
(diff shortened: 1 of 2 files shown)


entity_service_test.go
```diff
@@ -57,6 +57,7 @@ func TestEntityListOptions(t *testing.T) {
 		searchIDs           string
 		entityTypes         string
 		resolvePlaceholders bool
+		rendered            bool
```
nsiskind, line 60: I'd add a small comment on this line to be clear what this does.

nsiskind (approved):
lgtm

PR #277, opened 2022-07-15
making a user's createdDate field accessible

NathanielGE:
Created Date field added to User: Get here--> https://hitchhikers.yext.com/docs/knowledgeapis/accountsettings/users/users/


nsiskind (approved):
lgtm

PR #288, opened 2022-09-23
change isclosed from **bool to bool to treat nil and false the same

cdworak:


entity_diff_test.go
```diff
@@ -776,6 +776,99 @@ func TestEntityDiff(t *testing.T) {
 			newNilIsEmpty:  true,
 			isDiff:         false,
 		},
+		diffTest{
+			name:               "Hours: base is IsClosed: nil, new is IsClosed: false. Should not have diff because will result in API error ",
```
Zavoky: this name could be updated since isClosed can no longer be nil

entity_diff_test.go
```diff
@@ -776,6 +776,99 @@ func TestEntityDiff(t *testing.T) {
...
+					},
+					IsClosed: false,
+				}),
+				Sunday: NullableDayHours(&DayHours{
+					OpenIntervals: &[]Interval{
+						Interval{Start: "08:00", End: "16:30"},
+					},
+					IsClosed: false,
+				})}),
+			////newNilIsEmpty:      true,
```
nsiskind: delete comments?

nsiskind (approved):
lgtm

PR #292, opened 2023-01-13
modify AnalyticsData struct

mtusman:


analytics_data.go
```diff
@@ -25,9 +25,11 @@ type AnalyticsData struct {
 	GooglePhoneCalls                      *int     `json:"Google Phone Calls"`
 	YelpCustomerActions                   *int     `json:"Yelp Customer Actions"`
 	AverageRating                         *float64 `json:"Average Rating"`
-	NewReviews                            *int     `json:"Reviews"`
+	NewReviews                            *int     `json:"Reviews"`     // used to unmarshall new reviews for api versions on or after 20210729
+	NewReviewsAlternative                 *int     `json:"New Reviews"` // used to unmarshal new reviews for api versions before 20210729
 	StorepagesSessions                    *int     `json:"Storepages Sessions"`
-	StorepagesPageviews                   *int     `json:"Pages Views"`
+	StorepagesPageviews                   *int     `json:"Pages Views"` // used to unmarshall page views for api versions before 20210729
+	StorepagesPageviewsAlternative        *int     `json:"Page Views"`  // used to unmarshall new reviews for api versions on or after 20210729
```
nsiskind: nit but i think unmarshal only has 1 L

also should this comment say "page views" not "new reviews"?

lastly, the json tag "Pages Views" is repeated here. if the tag doesn't change, we don't need a new field. Otherwise this needs to be updated.

mtusman: Just updated the comment. The json tags are a bit different. One is `Pages Views` and the other is `Page Views` (without the s). Let me know if i missed something tho!

nsiskind (approved):
lgtm!

PR #296, opened 2023-04-13
add rankTrackingCompetitors to location entity

matthewleifer:


matthewleifer:
Entity with this field filled out can be seen here: https://www.yext.com/pagesadmin/entitydata?entityId=92187724&locale=

cblair28 (approved):
lgtm

PR #298, opened 2023-04-27
feat(custom): add GoogleEntityRelationship field/struct

kyled526:
to entity healthcare professional/healthcare facility for ETL use

J=PC-213266
TEST=manual

### DIFF
--- a/entity_healthcare_facility.go
+++ b/entity_healthcare_facility.go
@@ -8,6 +8,7 @@ const ENTITYTYPE_HEALTHCAREFACILITY EntityType = "healthcareFacility"
 
 // Location is the representation of a Location in Yext Location Manager.
 // For details see https://www.yext.com/support/platform-api/#Administration_API/Locations.htm
+
 type HealthcareFacilityEntity struct {
 	BaseEntity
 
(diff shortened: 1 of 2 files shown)


entity_healthcare_facility.go
```diff
@@ -110,6 +116,11 @@ type HealthcareFacilityEntity struct {
 	FirstPartyReviewPage *string `json:"firstPartyReviewPage,omitempty"`
 }
 
+type GoogleEntityRelationship struct {
```
AidanFitzgerald: you don't need to add this actually, we can just use the one already defined in entity_location.go! go will just find the struct since they're in the same package

entity_healthcare_facility.go
```diff
@@ -123,6 +134,15 @@ func (l *HealthcareFacilityEntity) UnmarshalJSON(data []byte) error {
 	return UnmarshalEntityJSON(l, data)
 }
 
+func NullableGoogleEntityRelationship(g *GoogleEntityRelationship) **GoogleEntityRelationship {
```
AidanFitzgerald: same w/ the helper funcs :)

entity_healthcare_professional.go
```diff
@@ -12,14 +12,16 @@ import (
...
-	GENDER_UNSPECIFIED                           = "UNSPECIFIED"
+	EDUCATION_FELLOWSHIP                                    = "FELLOWSHIP"
+	EDUCATION_INTERNSHIP                                    = "INTERNSHIP"
+	EDUCATION_MEDICAL_SCHOOL                                = "MEDICAL_SCHOOL"
+	EDUCATION_RESIDENCY                                     = "RESIDENCY"
+	ENTITYTYPE_HEALTHCAREPROFESSIONAL            EntityType = "healthcareProfessional"
+	GENDER_MALE                                             = "MALE"
+	GENDER_FEMALE                                           = "FEMALE"
+	GENDER_UNSPECIFIED                                      = "UNSPECIFIED"
+	GOOGLE_ENTITY_RELATIONSHIP_TYPE_DEPARTMENTIN            = "DEPARTMENT_IN"
```
AidanFitzgerald: and same here haha, you prob can just add them in one place and then you don't need to redeclare them. love that you're adding the consts tho!!

AidanFitzgerald (approved):

PR #301, opened 2023-09-08
migrate custom field endpoint to optionally use the configuration end…

mtusman:
…point

J=CR-3126
TEST=auto

### DIFF
--- a/cac.go
+++ b/cac.go
@@ -1,0 +1,300 @@
+package yext
+
+import "unicode"
+
+const (
+	CONFIGFIELDTYPE_YESNO           = "boolean"
+	CONFIGFIELDTYPE_DECIMAL         = "decimal"
+	CONFIGFIELDTYPE_DATE            = "date"
+	CONFIGFIELDTYPE_DAILYTIMES      = "dailyTimes"
+	CONFIGFIELDTYPE_HOURS           = "hours"
+	CONFIGFIELDTYPE_IMAGE           = "image"
+	CONFIGFIELDTYPE_OPTION          = "option"
+	CONFIGFIELDTYPE_VIDEO           = "video"
+	CONFIGFIELDTYPE_RICHTEXT        = "richText"
+	CONFIGFIELDTYPE_RICHTEXTV2      = "richTextV2"
+	CONFIGFIELDTYPE_LIST            = "list"
+	CONFIGFIELDTYPE_STRING          = "string"
+	CONFIGFIELDTYPE_MARKDOWN        = "markdown"
+	CONFIGFIELDTYPE_CTA             = "cta"
+	CONFIGFIELDTYPE_PRICE           = "price"
+	CONFIGFIELDTYPE_ENTITYREFERENCE = "entityReference"
+)
+
+type ConfigFieldEligibilityGroup struct {
+	Id         *string    `json:"$id,omitempty"`
+	Schema     string     `json:"$schema,omitempty"`
+	Name       string     `json:"name,omitempty"`
+	EntityType EntityType `json:"entityType,omitempty"`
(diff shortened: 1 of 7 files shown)


cac.go
```diff
@@ -0,0 +1,300 @@
...
+		Group:                      "NONE",
+		Description:                c.Description,
+		EntityAvailability:         entityAvailability,
+	}
+
+	if c.Localization == "LOCALE_SPECIFIC" {
+		r.AlternateLanguageBehaviour = "LANGUAGE_SPECIFIC"
+	}
+
+	for i, char := range c.Group {
```
bhainesva, line 180: I think this logic will have issues with 2-digit groups, "GROUP10" for ex

cac_service.go
```diff
@@ -0,0 +1,134 @@
...
+			return r, err
+		}
+	}
+
+	return r, nil
+}
+
+func (s *ConfigFieldService) Edit(cf *CustomField) (*Response, error) {
+	updatedDefinition := cf.TransformToConfigField()
+	if cf.Type == CUSTOMFIELDTYPE_SINGLEOPTION || cf.Type == CUSTOMFIELDTYPE_MULTIOPTION {
```
bhainesva, line 87: Are the validation settings only relevant for option fields?

cac_service.go
```diff
@@ -0,0 +1,134 @@
...
+		if cf.Type == CUSTOMFIELDTYPE_SINGLEOPTION {
+			updatedDefinition.Type.OptionType.Option = cf.GetConfigOptions()
+		} else {
+			updatedDefinition.Type.ListType.Type.ConfigType.OptionType.Option = cf.GetConfigOptions()
+		}
+	} else {
+		updatedDefinition.Type = nil
+	}
+
+	r, err := s.client.DoRequestJSON("PATCH", fmt.Sprintf("%s/%s/%s", ConfigResourcesPath, ConfigFieldSubType, GetString(cf.Id)), cf.TransformToConfigField(), nil)
```
bhainesva: We've potentially made some modifications to `updatedDefinition` at this point, is it ok that we're using `cf.TransformToConfigField()` here and not `updatedDefinition`?

customfield.go
```diff
@@ -105,6 +118,247 @@ func (c *CustomField) UnmarshalJSON(data []byte) error {
...
+	return nil
+}
+
+func (c *CustomField) GetConfigOptions() []ConfigOption {
+	var (
+		r []ConfigOption
+	)
+
+	configFieldTypeId := c.GetConfigFieldTypeId()
+	if !(configFieldTypeId == CONFIGFIELDTYPE_OPTION || configFieldTypeId == CONFIGFIELDTYPE_LIST) {
```
bhainesva, line 340: Mostly out of curiosity, is this check necessary or would the c.Options loop just do nothing for invalid field types? Do we need to distinguish between multioption lists and other lists?

cac_service.go
mtusman: Yes, I couldn't find any modifications / updates to the validation settings by an existing ETL using the custom field endpoint. Most modifications are related to updating single option or multi options and so I wanted to make sure that when that happens, we don't override any existing CF validations.

customfield.go
mtusman: Don't think it matters. I just added it because un-marshalling the existing custom field endpoint returns nil and not an empty list

cac_service.go
mtusman: Whoops, this is an oversight from my end.

cac.go
mtusman: Ah yes, forgot about that.

bhainesva (approved):

PR #325, opened 2024-10-08
feat(healthcare-professional): Add YextDisplayCoordinate property and…

tritpham:
… Get functions


entity_healthcare_professional.go
```diff
@@ -425,6 +426,22 @@ func (y HealthcareProfessionalEntity) GetRoutableLng() float64 {
 	return 0
 }
 
+func (y HealthcareProfessionalEntity) GetYextDisplayLat() float64 {
```
tritpham, line 429: Similar to entity_location.go https://github.com/yext/yext-go/blob/entities/entity_location.go#L842

stephh2 (approved):

PR #329, opened 2025-01-22
fix(custom): add option for list CF types and proper fallback for CFT…

Jeffrey-Rhoads17:
…s to set the type correctly

J=PC-270239
TEST=manual


jknutsonn (approved):
LGTM

